PR #6, opened 2016-02-17
BUG: don't exit in package init for version check

crosbymichael:
It is horrible to have a package exit your application in its init
without the consumer being able to do anything about it.  It is better
to return and error if the lib version is too low and have the consumer
of your package decide what to do instead of exit their application.
This changes it to return a typed error if the lib is to low whenever
you try to create a filter instead of doing this check in init()

Signed-off-by: Michael Crosby crosbymichael@gmail.com

### DIFF
--- a/seccomp.go
+++ b/seccomp.go
@@ -9,6 +9,7 @@
 package seccomp
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"runtime"
@@ -26,6 +27,10 @@ import (
 import "C"
 
 // Exported types
+
+// ErrUnsupportedVersion is returned when the libseccomp version on the system is to low for use
+// with the go bindings.
+var ErrUnsupportedVersion = errors.New("Libseccomp version too low: minimum supported is 2.1.0")
 
 // ScmpArch represents a CPU architecture. Seccomp can restrict syscalls on a
 // per-architecture basis.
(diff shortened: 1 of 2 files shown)


mheon:
I strongly suspect that the bindings won't even compile with libseccomp < 2.1.0 due to missing definitions in the header - I'd be interested in hearing if there's an exception to that?

Otherwise: I think that this is a reasonable change, though perhaps the init() block should be converted into logging a warning instead of being removed entirely? I'd just like to make it clear that the bindings are untested with < 2.1.0.

crosbymichael:
Because the C libs are dynamically linked it can compile but then not run on another system.  

Logging in init() or in libs is also pretty bad because you get output in your application that you cannot control.  The consumers can use the lib to check the version and they will also get an error that they can handle how they please.

pcmoore:
FWIW, I also plan on adding an API to libseccomp v2.3.0 that will allow the golang bindings to dynamically query the library for version information.

* https://github.com/seccomp/libseccomp/issues/5

mheon:
@crosbymichael I don't think this will fix the issue with dynamically-linked libseccomp, given that we resolve the symbols for library version at compile time. This check should never trigger if you dynamic link in a version <2.1.0 as long as it was originally compiled with a supported version.

mheon:
So, with the recent release of libseccomp 2.3.0, we now have a way of obtaining the version of libseccomp we're linked against at runtime, which solves the previous issue. I'll be working on an update for the Go bindings over the next few days, and I'll see about incorporating these changes in there.

crosbymichael:
thx

pcmoore:
@mheon Any updates you can share?

pcmoore:
@mheon Me again, being annoying ... thoughts on this?

crosbymichael:
I think the change still applies, you do not want to exit programs in your package init, return an error and let the caller decide what to do, so this has very little about linking.

pcmoore:
Closing this out as it has been replaced by PR #16.

PR #8, opened 2016-03-21
seccomp: use pkg-config for cgo flag generation

cyphar:
Not all distributions package libseccomp in the same way, but pkg-config
allows the same configuration to work on different distributions. Switch
to using pkg-config to automatically figure out what the correct
commandline flags are for libseccomp.

Signed-off-by: Aleksa Sarai <asarai@suse.de>

### DIFF
--- a/seccomp.go
+++ b/seccomp.go
@@ -20,7 +20,7 @@ import (
 
 // C wrapping code
 
-// #cgo LDFLAGS: -lseccomp
+// #cgo pkg-config: libseccomp
 // #include <stdlib.h>
 // #include <seccomp.h>
 import "C"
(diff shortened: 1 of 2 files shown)


mheon:
This seems entirely reasonable. I'd be glad to get this in, pending a few tests to make sure nothing broke.

pcmoore:
Thanks for the patch.  This looks good to me, I'll merge this once I fix the libseccomp v2.3.0 issue on 32-bit x86.

pcmoore:
Merged in 38efe4edc93552bfb879ca93b7b5e161f205d714, thanks.

Issue #7, opened 2016-03-11
RFE: add support for MIPS/PPC/s390 to the golang bindings


pcmoore:
As requested by Phil Estes along with a patch:

 * https://groups.google.com/forum/#!topic/libseccomp/JXRvNsdGGPw

estesp:
Now that it looks like the main C lib is up to v2.3.1, is there any chance some of these changes can get into the Go bindings and a release can be cut?  Thanks!

pcmoore:
Merged in 60c9953736798c4a04e90d0f3da2f933d44fd4c4, thanks for your patience!

michael-holzheu:
I am no expert for libseccomp-golang, but aren't changes for `seccomp_internal.go` missing to be complete?

Something like the following (untested) patch: 2e378814b62dc4af74259914a29c5621bb91c744

michael-holzheu:
In case I am right, for your convenience, I created PR #10.

michael-holzheu:
ping @tophj-ibm

pcmoore:
PR #10 merged via 32f571b70023028bd57d9288c20efbcb237f3ce0.

PR #10, opened 2016-05-31
BUG: add missing defines for PPC / S390 goloang bindings

michael-holzheu:
Add missing PPC and s390 code for issue #7 to file "seccomp_internal.go".

Signed-off-by: Michael Holzheu <holzheu@linux.vnet.ibm.com>

### DIFF
--- a/seccomp_internal.go
+++ b/seccomp_internal.go
@@ -58,6 +58,26 @@ const uint32_t C_ARCH_BAD = ARCH_BAD;
 #define SCMP_ARCH_MIPSEL64N32 ARCH_BAD
 #endif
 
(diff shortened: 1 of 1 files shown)


estesp:
Good catch--I think this is definitely correct, but usually the libseccomp code is managed not via GH PRs; you may have to submit via the mailing list for @pcmoore to handle.

pcmoore:
Thanks, merged via 32f571b70023028bd57d9288c20efbcb237f3ce0.

@estesp You are correct, generally I prefer patches to be submitted via the mailing list as you describe, but I don't ignore PRs.

PR #12, opened 2016-08-25
RFE: split into small files for much better readability

ProhtMeyhet:


pcmoore:
Thank you for your contribution but this is not a chance I believe in making at this time.

PR #13, opened 2016-08-25
RFE: add some examples

ProhtMeyhet:
this pull request is independent from my previous pull request #12 

tested on x86_64. i haven’t had time to set up go on my arm6 raspi yet. please test, test on other architectures and have a look if i got the code and the comments correct.

missing is a multithreaded example for different threads, since it's not really in the go language to start a new thread.

### DIFF
--- a/examples/whitelist-demo/main.go
+++ b/examples/whitelist-demo/main.go
@@ -1,0 +1,110 @@
+package main
+
+// +build linux
+
+import(
+	// only used for output
+	"fmt"
+	// only used for os.Exit
+	"os"
+
+	seccomp "github.com/seccomp/libseccomp-golang"
+)
+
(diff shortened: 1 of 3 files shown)


pcmoore:
We actually have a large number of tests/examples in the main libseccomp repository, written in both C and Python.  Unfortunately due to way Go libraries are typically distributed it is difficult to maintain Go versions of the tests/examples in the same directory.

If you are interested, I think it would be worthwhile in porting some, or all, of those examples to Go.

pcmoore:
@ProhtMeyhet I'm going to close this PR, but I would be very happy to work with you to port of some the existing libseccomp tests to the libseccomp-golang bindings (see my comment above).  If you are interested in working on this, please let me know.

Issue #11, opened 2016-07-18
RFE: runtime version detection


mYmNeo:
From #938(https://github.com/opencontainers/runc/issues/938)
There are three variables called verMajor, verMinor, verMicro to identify the version of libseccomp library, but they are fixed after building.If we upgrading the libseccomp library, the binary can not detect the new version.Should we change the detection fix mode to runtime mode such as using pkg --modversion libseccomp

pcmoore:
@mYmNeo I have no idea what to what project #938 refers; I'm going to guess Docker or something similar; for reference please update this with a link to the project/issue.

Regarding the runtime version detection, we probably want to make use of the new seccomp_version() API in the base libseccomp library, we've talked a little about this in PR https://github.com/seccomp/libseccomp-golang/pull/6.

mheon:
The issue with using runtime detection, specifically in the case of Docker, is that it's a new function, and Docker wants to link against a great many versions of the library. Right now, the situation is definitely not ideal (Docker uses the featureset of whatever version of the library it compiles against, which can lead to passing unrecognized constants into the library - like, say, SCMP_ARCH_S390X into a v2.1.1 library which doesn't have the S390 patches). However, if we move to runtime detection, we make the problem worse. We're no longer passing an unknown constant into earlier library versions, and instead we're calling a function that does not exist. Link errors are less desirable than error returns from the library, so in a way it makes things worse. The real solution would be to require that you link against the version of libseccomp you built against, but that's not really an option due to package distribution issues.

Runtime detection of what version of the library we link against is still a desirable thing, and hopefully we'll be able to turn it on in the future. Or maybe we can get Docker in Debian building against the backported v2.2.x library instead of stable v2.1.1, which I believe is most of the justification for requiring linking against many versions.

If this isn't actually a Docker issue, then this will sound very silly indeed, and I apologize!

mYmNeo:
@mheon Yes, this issue is about a sub-project called opencontainers/runc which is related to Docker. I just suffered from the wrong version detection.:disappointed:

pcmoore:
Fixed via e929b20e6658475be1134f17e29d755c676823c6, if you notice any additional problems please re-open this issue.  Thanks!

Issue #4, opened 2015-11-15
Q: release tagging


jessfraz:
it looks nicer when we vendor into docker/docker, no rush just wondering :)

mheon:
Planning on cutting a new release of the bindings soon, to coincide with a planned release of libseccomp itself --- can probably tag the bindings once that happens. We do need to figure out a version format for the bindings, though - @pcmoore, thoughts on that?

pcmoore:
I'm focusing on getting some things fixed in the main libseccomp library first so we can get that release out the door.  See here: https://groups.google.com/forum/#!topic/libseccomp/v6Sx3rObG5A

My initial thinking is to keep the version format/numbering the same as the main libseccomp release to keep things consistent - any objections?

mheon:
Sounds good to me, though there might eventually be need for another version field if we need multiple versions of the bindings for a single library version (if, say, we had a bug in the 2.2.3 bindings and needed a new bindings release, I'd want to distinguish them as 2.2.3.1 or similar).

pcmoore:
Yep, sounds good to me.

pcmoore:
@mheon I've given this some more thought and wonder if matching the main libseccomp version numbers might be a mistake.  As you pointed out, it seems very likely that we will need to iterate the two packages independently of each other.  Any idea what other golang bindings do?

pcmoore:
@jessfraz Since you're the reporter here, any opinions on how best to handle versioning of golang projects such as this which are just bindings to libraries written in other languages?  How do most projects handle this today: independent version numbers or sync with the main project?

mheon:
Sorry for the delay here, I'll try and find some time to take a look into this today.

pcmoore:
No worries, it really is just a question regarding the version numbers for the golang bindings.  We've got most of the pressing issues sorted such that I'm comfortable tagging a release, I'm just waffling a bit on the most golang-y approach.

thorduri:
@pcmoore 

> I'm just waffling a bit on the most golang-y approach. 

Can you expand on this this point ?

My two cents:
Matching the `seccomp` versions feels strange off the top of my head as the bindings support multiple versions of `seccomp` -- and if the bindings need a bump (for whatever reason) it can
easily become messy.

pcmoore:
To be clear, when I say waffling, I mean waffling on the version number for the golang bindings (see my above comments).  I would prefer to stay consistent with other golang bindings (the reason for my "golang-y" comment above).

[...]

PR #15, opened 2016-11-23
RFE: introduce GetRuntimeLibraryVersion

thorduri:
Uses the libseccomp `seccomp_version` (v2.3.0) to retrive the runtime version of the library.

This should probably not be merged until the discussions in #11 and #4 are finished.

I'd like to advocate for the removal of the `init()` error and exit, it's unidiomatic.
Returning an error from `NewFilter` based off of the current release of the bindings
when that happens, and the runtime version (slight modification of #6).

Signed-off-by: Thordur Bjornsson <thorduri@secnorth.net>

### DIFF
--- a/seccomp.go
+++ b/seccomp.go
@@ -302,6 +302,14 @@ func GetLibraryVersion() (major, minor, micro int) {
 	return verMajor, verMinor, verMicro
 }
 
+// GetRuntimeLibraryVersion returns the runtime version of the library the
+// bindings are using.
+// The version is formatted as follows: Major.Minor.Micro
+func GetRuntimeLibraryVersion() (major, minor, micro int) {
+	ver := C.seccomp_version()
+	return int(ver.major), int(ver.minor), int(ver.micro)
+}
+
 // Syscall functions
 
 // GetName retrieves the name of a syscall from its number.
--- a/seccomp_test.go
+++ b/seccomp_test.go
@@ -152,6 +152,19 @@ func TestMakeCondition(t *testing.T) {
 	}
 }
 
+// General Utility Function Tests
+
+func TestGetRuntimeLibraryVersion(t *testing.T) {
+	// Under test, we assume the library we build and run with are the same.
+	bmajor, bminor, bmicro := GetLibraryVersion()
+	rmajor, rminor, rmicro := GetRuntimeLibraryVersion()
+
+	if bmajor != rmajor || bminor != rminor || bmicro != rmicro {
+		t.Errorf("Build library version should match runtime library version")
+	}
+}
+
+
 // Utility Function Tests
 
 func TestGetNativeArch(t *testing.T) {


mheon:
Looks good when built against versions of libseccomp supplying `seccomp_version`, but we can't guarantee we're building against v2.3.0. We could add gating on library version and return an error when the predefined version constants are less than v2.3.0. Alternatively, this could be combined with the existing `GetLibraryVersion` and use `seccomp_version` on versions 2.3.0 and higher, and otherwise use the preprocessor defines already used in `GetLibraryVersion.`

On the `init()` error - I tend to agree, though I'd prefer if we don't just check at `NewFilter()` but also all the non-filter-related functions that call into the library (for example `GetSyscallFromName`).

thorduri:
> Looks good when built against versions of libseccomp supplying seccomp_version, but we can't guarantee we're building against v2.3.0.

Apologies if I wasn't clear enough, or failed to communicate my assumptions. 

My hope is that when the discussions in #11 and #4 are concluded, and tagging parity is reached is
that version x.y.z of the bindings works with version x.y.z of the library, without any gating of functionality -- i.e. building version x.y.z of the binding, must be done against version x.y.z of the library, and the runtime must be x.y.z as well.
 
Is that reasonable ?

> On the init() error - I tend to agree, though I'd prefer if we don't just check at NewFilter() but also all the non-filter-related functions that call into the library (for example GetSyscallFromName).

Good point. I will do the `checkVersionAbove` dance where applicable.

mheon:
When the bindings were included in Docker, we received  a request from their maintainers that we allow a single Docker binary to be linked against multiple versions of libseccomp, which is why the version checks were originally included. I can ask if this is no longer a requirement on their end, but I suspect it still is.

The biggest issue in contention was support of multiple rules matching the same syscall, but with different arguments specified. This is broken in v2.1.x of libseccomp (bug in the filter generation code), and fixed in v2.2.x. I had/have a strong preference for not building at all on v2.1.x because it's a difficult to notice bug that can serious affect the integrity of a Seccomp profile, but Docker really wanted to be able to use libseccomp from the Debian repositories (which are still shipping v2.1.1 in Jessie). Given that Docker builds from the same source tree on platforms with v2.1.x of the library and v2.2.x, the checks needed to happen in the bindings to make this possible.

It's definitely not a good solution or an ideal situation right now. I don't think we can drop the version checks entirely, though it might be enough to shove them in a branch and keep them out of master.

thorduri:
@mheon Thanks a lot for the historical context, the situation makes much more sense now.

> It's definitely not a good solution or an ideal situation right now. I don't think we can drop the version checks entirely, though it might be enough to shove them in a branch and keep them out of master.

With that in mind and #16 it doesn't bother me at all, if anything it makes usage much simpler esp.
on LTS distributions.

Still want to see those tags though 😉

thorduri:
@pcmoore @mheon Worth keeping open ?

mheon:
@thorduri We have a library version function now, and API levels (another way of telling if the library supports a given function) will be coming in the next minor version release of the main library. I think this is safe to close.

Issue #17, opened 2016-12-02
RFE: set the SCMP_FLTATR_CTL_TSYNC by default in the golang bindings


thorduri:
Hi,

Due to the way the `go` runtime works (right off the bat on starts a handful of threads),
calling `SetTsync(true)` in `Load()` might reduce surprises for the package users.

Further, as a golang programmer, you have no control over thread creation / destruction or
scheduling (there are nuances here, but generally) so this doesn't map very well.

I'd go further, and suggest the API for setting/clearing this attribute should be removed
from the package and this just be done implicitly at Load time.

Prompted by https://github.com/opencontainers/runc/issues/1212

Comments ?

pcmoore:
This seems perfectly reasonable to me, and isn't likely to cause a problem with any existing libseccomp-golang users due to the thread awareness limitations of Go.  @mheon what do you think?

mheon:
General idea sounds fine here. You'd need to ignore potential error returns (library 2.1.x won't have the appropriate flag) when setting it in `Load` which is annoying --- it'd be nice if there were a way to identify that the flag was not set because of a too-early library version. Of course, we don't check if we're running on a kernel with `tsync` support anyways, so it's not that different from current behavior.

pcmoore:
@thorduri @mheon okay, who wants to code this up into a patch/PR? ;)

thorduri:
@pcmoore I can handle that.

However I think https://github.com/seccomp/libseccomp-golang/pull/15 would need to be merged first to address the issue pointed out by @mheon -- I've only verified libseccomp 2.3.1, however earlier versions might work (need to double check).

FWIW, docker et al, seem to "ship" 2.3.1, see e.g. [here](https://github.com/docker/docker/blob/master/Dockerfile#L112).

@mheon Do you think it would be worth it to gate this behaviour for kernels < 3.17 (which was released in Oct 2014 -- I don't) ?

mheon:
@thorduri I wouldn't bother gating on the kernel. Perhaps a note in the documentation somewhere wouldn't be amiss, though. A note that we automatically enable tsync, but only if kernel and libseccomp are of sufficient version.

pcmoore:
You can always just call seccomp_attr_set(SCMP_FLTATR_CTL_TSYNC) and if it succeeds then everything is okay, if not, well it's not supported :)

mheon:
[...]

PR #18, opened 2016-12-03
RFE: bump minimum required version to 2.2.0

thorduri:
Set the SCMP_FLTATR_CTL_TSYNC attribute in NewFilter, and error out should
that fail.

Since the threads are hidden from the programmer, it is inherently unsafe
to run without this bit set, as the runtime can and does create threads
very early, which are used in subsequent operations (go routines), meaning
that the filter might not be applied, see discussion in #17.

Signed-off-by: Thordur Bjornsson <thorduri@secnorth.net>

### DIFF
--- a/seccomp.go
+++ b/seccomp.go
@@ -429,7 +429,8 @@ type ScmpFilter struct {
 	lock      sync.Mutex
 }
 
-// NewFilter creates and returns a new filter context.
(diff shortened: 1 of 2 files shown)


thorduri:
@pcmoore @mheon Worth rebasing and keeping open ?

mheon:
@thorduri We set TSync by default now, so I think this is obsolete

PR #21, opened 2017-03-13
BUG: fix runtime version checking

### LINKED ISSUE #20: BUG: version checking does not work for executable linked against the shared library
For `go` binaries that uses shared library, e.g. `docker-runc`. The version checking really doesn't work. From the look of things, the version is determined at compile time via the header file `seccomp.h`. So, we have a situation where a [published Docker 1.13 RPM](http://download.opensuse.org/repositories/Virtualization:/containers/openSUSE_Leap_42.1/x86_64/docker-1.13.0-182.1.x86_64.rpm) probably is built against a version of `libseccomp` that's older than 2.2.1. The result is a run-time error:

> `Error response from daemon: oci runtime error: container_linux.go:247: starting container process caused "conditional filtering requires libseccomp version >= 2.2.1`

So the user upgraded `libseccomp` to 2.3 and the error remains. That's because the version check is done against the header values captured at build time. It'd seem that `libseccomp-golang` works correctly only if `libseccomp` is statically linked into the binary.

dvusboy:
* add three C functions to retrieve the major, minor and micro
  version values from `seccomp_version()` when that's available
  under v2.3.x or newer
* has to change type from `int` to `uint` as that's the type
  used in `struct scmp_version`

### DIFF
--- a/seccomp_internal.go
+++ b/seccomp_internal.go
@@ -120,9 +120,37 @@ const int      C_CMP_GE            = (int)SCMP_CMP_GE;
 const int      C_CMP_GT            = (int)SCMP_CMP_GT;
 const int      C_CMP_MASKED_EQ     = (int)SCMP_CMP_MASKED_EQ;
 
-const int      C_VERSION_MAJOR     = SCMP_VER_MAJOR;
-const int      C_VERSION_MINOR     = SCMP_VER_MINOR;
-const int      C_VERSION_MICRO     = SCMP_VER_MICRO;
+#if SCMP_VER_MAJOR == 2 && SCMP_VER_MINOR >= 3
+unsigned int get_major_version()
+{
+       return seccomp_version()->major;
+}
+
+unsigned int get_minor_version()
+{
+       return seccomp_version()->minor;
+}
+
+unsigned int get_micro_version()
+{
+       return seccomp_version()->micro;
+}
+#else
+unsigned int get_major_version()
+{
+       return (unsigned int)SCMP_MAJOR_VERSION;
+}
+
+unsigned int get_minor_version()
+{
+       return (unsigned int)SCMP_MINOR_VERSION;
+}
+
+unsigned int get_micro_version()
+{
+       return (unsigned int)SCMP_MICRO_VERSION;
+}
+#endif
 
 typedef struct scmp_arg_cmp* scmp_cast_t;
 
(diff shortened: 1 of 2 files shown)


seccomp_internal.go
```diff
@@ -177,15 +205,15 @@ var (
 	// Error thrown on bad filter context
 	errBadFilter = fmt.Errorf("filter is invalid or uninitialized")
 	// Constants representing library major, minor, and micro versions
-	verMajor = int(C.C_VERSION_MAJOR)
-	verMinor = int(C.C_VERSION_MINOR)
-	verMicro = int(C.C_VERSION_MICRO)
+	verMajor     = uint(C.get_major_version())
```
mheon: gofmt flags the whitespace on these three lines as incorrect - you can fix with gofmt -s -w

mheon:
One nit, but otherwise looks good. The API change is annoying but shouldn't break anything significant. We could cast back to uint if we really wanted to, but I don't know if it's even a good idea (I sincerely hope we never have to deal with negative version numbers).

Overall, looks fine. I like that it still retains the option of building against 2.2.x and earlier if available.

seccomp_internal.go
dvusboy: Updated with an amend. I'm sorry about the noob mistake.

mheon: Looks fine now, thanks!

mheon:
@pcmoore Mind taking a look at this when you get a chance?

dvusboy:
Let me know if you want to keep the type in the public interface as `int`. I do think matching the type in `libseccomp` makes more sense to me. But the API change is "annoying", as you said.

dvusboy:
BTW, I did file [a bug report](https://bugzilla.opensuse.org/show_bug.cgi?id=1029222) on the OpenSuSE RPM. The more I think about this, the more I believe the issue lies with the package maintainer rather than Docker. The one thing I think the Docker folks can do is to pick the right default profile given the version of `libseccomp`. But that obviously requires `seccomp_version()`...

pcmoore:
> @pcmoore Mind taking a look at this when you get a chance?

@mheon I'll leave the decision up to you, if you're happy with this PR that is fine with me.

mheon:
@pcmoore Sounds good. I think this is in good shape to merge, then. The public API change, while annoying, is more correct after the change - I wonder what past-me was thinking when I made version numbers signed. And I really doubt anything will be broken now that our version is in uints instead of ints. Hopefully I can find some time tonight to do final review + merge.

pcmoore:
No worries, we haven't tagged a v1.0.0 release yet so I don't think we need to worry too much about API breakage just yet.  Once we hit v1.0.0 we'll have to be much more careful.

mheon:
@dvusboy One more quick thing before I merge - would you mind adding a Signed-off-by to the commit? `git commit --amend -s` should do it.

dvusboy:
@mheon Any update?

mheon:
@dvusboy Oh, sorry, didn't get a notification you'd updated. I'll get this merged tonight. Thanks!

dvusboy:
Yeah, I was suspecting the thumbs-up was too subtle. Thank you for your help.

mheon:
Committed to master via 9814e55a2e59df39b8ad4fbff1226585ebb84674, thanks!

PR #26, opened 2017-05-30
RFE: introduce ErrSyscallDoesNotExists error type

mvo5:
A small trivial branch that makes the `could not resolve syscall name` its own error type.

This way I can write client code like this:
```
scmpSys, err := GetSyscallFromNameByArch(name, arch)
if err == ErrSyscallDoesNotExist {
    println("warning: skipping %q - not available on %s", name, arch)
} else if err != nil {
    return err
}
```

I.e. it allows me to differentiate an error that I want to ignore from a real error like unsupported architecture or an unsupported underlying library version.

Happy to adjust both the string and the ErrXXX name as you prefer of course.

### DIFF
--- a/seccomp.go
+++ b/seccomp.go
@@ -168,6 +168,10 @@ const (
 	CompareMaskedEqual ScmpCompareOp = iota
 )
 
+var (
+	ErrSyscallDoesNotExist = fmt.Errorf("could not resolve syscall name")
+)
+
 // Helpers for types
 
 // GetArchFromString returns an ScmpArch constant from a string representing an
@@ -350,7 +354,7 @@ func (s ScmpSyscall) GetNameByArch(arch ScmpArch) (string, error) {
 
 	cString := C.seccomp_syscall_resolve_num_arch(arch.toNative(), C.int(s))
 	if cString == nil {
-		return "", fmt.Errorf("could not resolve syscall name")
+		return "", ErrSyscallDoesNotExist
 	}
 	defer C.free(unsafe.Pointer(cString))
 
@@ -373,7 +377,7 @@ func GetSyscallFromName(name string) (ScmpSyscall, error) {
 
 	result := C.seccomp_syscall_resolve_name(cString)
 	if result == scmpError {
-		return 0, fmt.Errorf("could not resolve name to syscall")
+		return 0, ErrSyscallDoesNotExist
 	}
 
 	return ScmpSyscall(result), nil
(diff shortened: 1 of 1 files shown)


mheon:
I'll give this a look later today, but this looks like a nice, simple change that is a clear enhancement - thanks!

mheon:
On the whole, code looks fine. However, a note on your example code: that should work fine even if the syscall in question is not available on the architecture you're using. `libseccomp` will gracefully handle such cases by ignoring rules related to that syscall, to allow for easier creation of multiarchitecture filters. You should never get a failure to resolve a syscall if it exists in the kernel for some architecture (for example, resolving `getuid32` on x86-64 machines will succeed even though it does not exist for that architecture). The only times a failure to resolve error should occur is when the syscall does not exist for any supported kernel or architecture (for example, trying to resolve "qwerty" or similar). Of course, these functions can produce other errors (notably the underlying library version being too low).

mvo5:
Anything I can do to move this PR foward? Happy to resolve the conflict if there is a chance that this gets merged.

mheon:
@mvo5 This looks fine, so if you resolve the conflict, I'd be happy to merge.

pcmoore:
@mvo5 any update on an updated PR?

rhatdan:
@mheon Is this something we should take over and Complete?

pcmoore:
@rhatdan @mheon it has been over a year with no response from @mvo5, if you wanted to take this over I would have no objection.  Just please make sure to credit @mvo5 in the updated patch.

drakenclimber (approved):
Looks good to me.

mvo5:
Hey, sorry for the lack of reply. I'm very happy if someone else would finish this, I'm a bit short on time right now and would still love to see something like this available in libseccomp. If noone else has time I can try to get to it at one of the following weekends.

pcmoore:
Hi @mvo5, I just merged this via 043204c0228bd0e72bed689c3219e5ecd14c9f75; thank you for the patch and your patience.  

As the patch need to most rewritten/ported due to changes in the main branch, and the lack of a sign-off from you, I simply redid the patch by hand (it was small) and gave you credit both in the patch description and in the "Suggested-by:" tag.  If you have a problem with that, please let me know.

mvo5:
Hi @pcmoore ! Thank you so much and sorry that I did not get back to you earlier, I meant to but then was busy with $work+$life :/ But I'm super happy this is available now, thanks!

pcmoore:
No problem @mvo5, my apologies it took so long to end up in the tree :)

PR #27, opened 2017-06-09
golang: add more info to errors with fmt.Errorf()

nodakai:
Didn't test any further than `make check`, so I'm not 100% confident that all the `Errorf()` invocations are soundly typed...

### DIFF
--- a/seccomp.go
+++ b/seccomp.go
@@ -211,7 +211,7 @@ func GetArchFromString(arch string) (ScmpArch, error) {
 	case "s390x":
 		return ArchS390X, nil
 	default:
-		return ArchInvalid, fmt.Errorf("cannot convert unrecognized string %s", arch)
+		return ArchInvalid, fmt.Errorf("cannot convert unrecognized string %q", arch)
 	}
 }
 
@@ -255,7 +255,7 @@ func (a ScmpArch) String() string {
 	case ArchInvalid:
 		return "Invalid architecture"
 	default:
-		return "Unknown architecture"
+		return fmt.Sprintf("Unknown architecture %#x", uint(a))
 	}
 }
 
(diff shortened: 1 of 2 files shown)


mheon:
After a brief check, looks fine other than potential format specifier issues. I'll try and get this reviewed/tested and merged later today.

seccomp.go
```diff
@@ -397,7 +397,7 @@ func GetSyscallFromNameByArch(name string, arch ScmpArch) (ScmpSyscall, error) {
 
 	result := C.seccomp_syscall_resolve_name_arch(arch.toNative(), cString)
 	if result == scmpError {
-		return 0, fmt.Errorf("could not resolve name to syscall")
+		return 0, fmt.Errorf("could not resolve name to syscall: %q on %#x", name, uint(arch))
```
mheon: Should use arch.String() here instead of uint(arch) - much more useful to get an error with a human-readable architecture

seccomp_internal.go
```diff
@@ -319,11 +319,11 @@ func (f *ScmpFilter) addRuleGeneric(call ScmpSyscall, action ScmpAction, exact b
 // Helper - Sanitize Arch token input
 func sanitizeArch(in ScmpArch) error {
 	if in < archStart || in > archEnd {
-		return fmt.Errorf("unrecognized architecture")
+		return fmt.Errorf("unrecognized architecture %#x", uint(in))
 	}
 
 	if in.toNative() == C.C_ARCH_BAD {
-		return fmt.Errorf("architecture is not supported on this version of the library")
+		return fmt.Errorf("architecture %#x is not supported on this version of the library", uint(in))
```
mheon: Since we've already verified that we recognize the arch above, can use arch.String() to print it here as well

mheon:
Two nits on printing human-readable strings instead of ints, otherwise looks good.

nodakai:
Thank you for the comments. Updated the patch accordingly

nodakai:
Ping

mheon:
@nodakai Sorry for the delay, committed to master as f6ec81daf48e41bf48b475afc7fe06a26bfb72d1

PR #29, opened 2017-09-21
expose improved kernel logging features through libseccomp-golang

tyhicks:
This is a pull request that complements https://github.com/seccomp/libseccomp/pull/92. It is not yet ready for merging as there's some pending work left to do for that libseccomp PR to allow applications, such as the libseccomp and libseccomp-golang test suites, to request/force a certain API level. See https://github.com/seccomp/libseccomp/pull/92#discussion_r138680169 for more context.

The SECCOMP_FILTER_FLAG_LOG filter flag and SECCOMP_RET_LOG action allow applications to have more control over what actions should be logged.

The tests added by this PR pass when run under a kernel that has support for the new filter flag and action and a libseccomp that has been built with https://github.com/seccomp/libseccomp/pull/92. They fail otherwise, pending the API level changes that would allow the libseccomp-golang tests to request a certain API level.

### DIFF
--- a/seccomp.go
+++ b/seccomp.go
@@ -137,6 +137,10 @@ const (
 	ActTrace ScmpAction = iota
 	// ActAllow permits the syscall to continue execution
 	ActAllow ScmpAction = iota
+	// ActLog permits the syscall to continue execution after logging it.
+	// This action is only usable when libseccomp API level 3 or higher is
+	// supported.
+	ActLog ScmpAction = iota
 )
 
 const (
@@ -295,6 +299,8 @@ func (a ScmpAction) String() string {
 	case ActTrace:
 		return fmt.Sprintf("Action: Notify tracing processes with code %d",
 			(a >> 16))
+	case ActLog:
+		return "Action: Log system call"
 	case ActAllow:
 		return "Action: Allow system call"
 	default:
--- a/seccomp.go
+++ b/seccomp.go
@@ -749,6 +749,30 @@ func (f *ScmpFilter) GetNoNewPrivsBit() (bool, error) {
 	return true, nil
 }
 
+// GetLogBit returns the current state the Log bit will be set to on the filter
+// being loaded, or an error if an issue was encountered retrieving the value.
+// The Log bit tells the kernel that all actions taken by the filter, with the
+// exception of ActAllow, should be logged.
+// The Log bit is only usable when libseccomp API level 3 or higher is
+// supported.
+func (f *ScmpFilter) GetLogBit() (bool, error) {
+	log, err := f.getFilterAttr(filterAttrLog)
+	if err != nil {
+		api, apiErr := getApi()
+		if (apiErr != nil && api == 0) || (apiErr == nil && api < 3) {
+			return false, fmt.Errorf("getting the log bit is only supported in libseccomp 2.4.0 and newer with API level 3 or higher")
+		}
+
+		return false, err
+	}
+
+	if log == 0 {
+		return false, nil
+	}
+
+	return true, nil
+}
+
 // SetBadArchAction sets the default action taken on a syscall for an
 // architecture not in the filter, or an error if an issue was encountered
 // setting the value.
@@ -773,6 +797,28 @@ func (f *ScmpFilter) SetNoNewPrivsBit(state bool) error {
 	}
 
 	return f.setFilterAttr(filterAttrNNP, toSet)
+}
+
+// SetLogBit sets the state of the Log bit, which will be applied on filter
+// load, or an error if an issue was encountered setting the value.
+// The Log bit is only usable when libseccomp API level 3 or higher is
+// supported.
+func (f *ScmpFilter) SetLogBit(state bool) error {
+	var toSet C.uint32_t = 0x0
+
+	if state {
+		toSet = 0x1
+	}
+
+	err := f.setFilterAttr(filterAttrLog, toSet)
+	if err != nil {
+		api, apiErr := getApi()
+		if (apiErr != nil && api == 0) || (apiErr == nil && api < 3) {
+			return fmt.Errorf("setting the log bit is only supported in libseccomp 2.4.0 and newer with API level 3 or higher")
+		}
+	}
+
+	return err
 }
 
 // SetSyscallPriority sets a syscall's priority.
--- a/seccomp.go
+++ b/seccomp.go
@@ -328,6 +328,25 @@ func GetLibraryVersion() (major, minor, micro uint) {
 	return verMajor, verMinor, verMicro
 }
 
+// GetApi returns the API level supported by the system.
+// Returns a positive int containing the API level, or 0 with an error if the
+// API level could not be detected due to the library being older than v2.4.0.
+// See the seccomp_api_get(3) man page for details on available API levels:
+// https://github.com/seccomp/libseccomp/blob/master/doc/man/man3/seccomp_api_get.3
+func GetApi() (uint, error) {
+	return getApi()
+}
+
+// SetApi forcibly sets the API level. General use of this function is strongly
+// discouraged.
+// Returns an error if the API level could not be set. An error is always
+// returned if the library is older than v2.4.0
+// See the seccomp_api_get(3) man page for details on available API levels:
+// https://github.com/seccomp/libseccomp/blob/master/doc/man/man3/seccomp_api_get.3
+func SetApi(api uint) error {
+	return setApi(api)
+}
+
 // Syscall functions
 
 // GetName retrieves the name of a syscall from its number.
--- a/seccomp_internal.go
+++ b/seccomp_internal.go
@@ -74,10 +74,18 @@ const uint32_t C_ACT_ERRNO         = SCMP_ACT_ERRNO(0);
 const uint32_t C_ACT_TRACE         = SCMP_ACT_TRACE(0);
 const uint32_t C_ACT_ALLOW         = SCMP_ACT_ALLOW;
 
+// The libseccomp SCMP_FLTATR_CTL_LOG member of the scmp_filter_attr enum was
+// added in v2.4.0
+#if (SCMP_VER_MAJOR < 2) || \
+    (SCMP_VER_MAJOR == 2 && SCMP_VER_MINOR < 4)
+#define SCMP_FLTATR_CTL_LOG _SCMP_FLTATR_MIN
+#endif
+
 const uint32_t C_ATTRIBUTE_DEFAULT = (uint32_t)SCMP_FLTATR_ACT_DEFAULT;
 const uint32_t C_ATTRIBUTE_BADARCH = (uint32_t)SCMP_FLTATR_ACT_BADARCH;
 const uint32_t C_ATTRIBUTE_NNP     = (uint32_t)SCMP_FLTATR_CTL_NNP;
 const uint32_t C_ATTRIBUTE_TSYNC   = (uint32_t)SCMP_FLTATR_CTL_TSYNC;
+const uint32_t C_ATTRIBUTE_LOG     = (uint32_t)SCMP_FLTATR_CTL_LOG;
 
 const int      C_CMP_NE            = (int)SCMP_CMP_NE;
 const int      C_CMP_LT            = (int)SCMP_CMP_LT;
@@ -179,6 +187,7 @@ const (
 	filterAttrActBadArch scmpFilterAttr = iota
 	filterAttrNNP        scmpFilterAttr = iota
 	filterAttrTsync      scmpFilterAttr = iota
+	filterAttrLog        scmpFilterAttr = iota
 )
 
 const (
@@ -545,6 +554,8 @@ func (a scmpFilterAttr) toNative() uint32 {
 		return uint32(C.C_ATTRIBUTE_NNP)
 	case filterAttrTsync:
 		return uint32(C.C_ATTRIBUTE_TSYNC)
+	case filterAttrLog:
+		return uint32(C.C_ATTRIBUTE_LOG)
 	default:
 		return 0x0
 	}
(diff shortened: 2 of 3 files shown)


seccomp_internal.go
```diff
@@ -67,16 +67,26 @@ const uint32_t C_ARCH_PPC64LE      = SCMP_ARCH_PPC64LE;
 const uint32_t C_ARCH_S390         = SCMP_ARCH_S390;
 const uint32_t C_ARCH_S390X        = SCMP_ARCH_S390X;
 
+#ifndef SCMP_ACT_LOG
+#define SCMP_ACT_LOG 0x7ffc0000U
+#endif
```
tyhicks: Is this the right thing to do here? This is to allow a new libseccomp-golang to build with an older libseccomp which doesn't have SCMP_ACT_LOG support. Is it acceptable to fail the build if the installed libseccomp is not new enough or should I leave this #define here?

seccomp_internal.go
```diff
@@ -67,16 +67,26 @@ const uint32_t C_ARCH_PPC64LE      = SCMP_ARCH_PPC64LE;
...
 const uint32_t C_ACT_KILL          = SCMP_ACT_KILL;
 const uint32_t C_ACT_TRAP          = SCMP_ACT_TRAP;
 const uint32_t C_ACT_ERRNO         = SCMP_ACT_ERRNO(0);
 const uint32_t C_ACT_TRACE         = SCMP_ACT_TRACE(0);
+const uint32_t C_ACT_LOG           = SCMP_ACT_LOG;
 const uint32_t C_ACT_ALLOW         = SCMP_ACT_ALLOW;
 
+#ifndef SCMP_FLTATR_CTL_LOG
+#define SCMP_FLTATR_CTL_LOG _SCMP_FLTATR_MIN
+#endif
```
tyhicks: Similar question here regarding what to do if the installed libseccomp is not new enough to know about SCMP_FLTATR_CTL_LOG.

seccomp.go
```diff
@@ -730,6 +734,23 @@ func (f *ScmpFilter) GetNoNewPrivsBit() (bool, error) {
 	return true, nil
 }
 
+// GetLogBit returns the current state the Log bit will be set to on the filter
+// being loaded, or an error if an issue was encountered retrieving the value.
+// The Log bit tells the kernel that all actions taken by the filter, with the
+// exception of ActAllow, should be logged.
```
mheon: Can you add something in the comment regarding a minimum required kernel version for this working?

seccomp.go
```diff
@@ -756,6 +777,18 @@ func (f *ScmpFilter) SetNoNewPrivsBit(state bool) error {
 	return f.setFilterAttr(filterAttrNNP, toSet)
 }
 
+// SetLogBit sets the state of the Log bit, which will be applied on filter
+// load, or an error if an issue was encountered setting the value.
```
mheon: Same as above, please add something about minimum kernel version where this will take effect. And/or minimum libseccomp version.

seccomp_internal.go
mheon: This matches what we're been doing elsewhere - we need to support building against older libseccomp, so defining if it's undefined is necessary. So yes, this is the correct thing to do, though it's not very convenient or good practice in general.

seccomp_internal.go
mheon: This should also be correct.

seccomp_test.go
```diff
@@ -519,3 +531,64 @@ func TestRuleAddAndLoad(t *testing.T) {
 		t.Errorf("Syscall returned incorrect error code - likely not blocked by Seccomp!")
 	}
 }
+
+func TestLogAct(t *testing.T) {
```
mheon: Can we skip this if the libseccomp version we build against is too low to have support? We have a function to grab library version we built against, it should be possible to skip the test if it's detected as too low.

mheon:
Have a few minor concerns mostly related to tests/comments, but overall looks fine.

seccomp.go
tyhicks: Yes, this is a good idea. There's, unfortunately, now a little uncertainty around which kernel release it'll be present in due to an unrelated [issue](https://lkml.kernel.org/r/<alpine.LRH.2.21.1709101406320.22614@namei.org>). I'll update the comment as soon as there's some clarity.

seccomp.go
tyhicks: Sounds good.

seccomp_test.go
tyhicks: Yes! Good idea. Unfortunately, it doesn't solve the case where libseccomp is "new" but the kernel is "old" but at least it improves the situation.

pcmoore:
@tyhicks thanks for doing this.

@mheon thanks for reviewing, but please hold off on merging anything until we sort things out in the main repo.

mheon:
@pcmoore Roger that, was going to ask about that. I'll mark this as WIP until we get support merged into the main library.

tyhicks:
@mheon @pcmoore Hello! I've updated this PR with API level bindings (in support of https://github.com/seccomp/libseccomp/commit/e89d18205c7dcd7582f41051cd6389c9b12dfccf) and I've updated the SCMP_FLTATR_CTL_LOG and SCMP_ACT_LOG related patches accordingly for API level support and to match what is nearly merged in https://github.com/seccomp/libseccomp/pull/92.

seccomp_test.go
tyhicks: The new API level support in libseccomp allows us to handle the case that I mentioned above. The updated PR now handles all combinations of new/old libseccomp-golang, new/old libseccomp, and new/old kernel.

seccomp.go
tyhicks: In the update to this PR, I've added a comment about the minimum API level required (3).

seccomp.go
tyhicks: In the update to this PR, I've added a comment about the minimum API level required (3).

seccomp_internal.go
tyhicks: This ended up not working since SCMP_FLTATR_CTL_LOG is a member of an enum and the C preprocessor isn't able to handle enum members in ifndef conditionals. In the update to this PR, I conditionalized based on the (upcoming) libseccomp version of 2.3.3.

seccomp.go
```diff
@@ -135,6 +135,10 @@ const (
 	// ActTrace causes the syscall to notify tracing processes with the
 	// given error code. This code can be set with the SetReturnCode method
 	ActTrace ScmpAction = iota
+	// ActLog permits the syscall to continue execution after logging it.
+	// This action is only usable when libseccomp API level 3 or higher is
+	// supported.
+	ActLog ScmpAction = iota
```
mheon: Can you put this after ActAllow? By placing it between ActTrace and ActAllow, ActAllow's constant is renumbered versus current versions of the library

seccomp.go
```diff
@@ -328,6 +334,21 @@ func GetLibraryVersion() (major, minor, micro uint) {
 	return verMajor, verMinor, verMicro
 }
 
+// GetApi returns the API level supported by the system.
```
mheon: Can you give a brief description of what API levels there are in the comments for this and SetAPI?

mheon:
Can you add API level and/or library version checks in `getFilterAttr()` and `setFilterAttr()` for the new filterAttrLog? The library should catch it already, but it'd be nice to give a proper string error ("Setting the log bit is only supported in Libseccomp 2.3.3 with API level 3 set" or something similar). Doing a getAPI() call and returning an error if the API level call is not supported or the API level is too low should be sufficient.

seccomp.go
tyhicks: I sure can. My reasoning for not including API level descriptions in libseccomp-golang was that they may become outdated because developers could easily forget to keep the descriptions updated when new API levels are added. The `seccomp_api_get` man page is probably going to be the most authoritative location for API level descriptions but I don't know how you feel about pointing to that man page from the function documentation.

seccomp.go
```diff
@@ -328,6 +334,21 @@ func GetLibraryVersion() (major, minor, micro uint) {
...
+func GetApi() (uint, error) {
+	return getApi()
+}
+
+// SetApi forcibly sets the API level. General use of this function is strongly
+// discouraged.
+// Returns an error if the API level could not be set. An error is always
+// returned if the library is older then v2.3.3
+func SetApi(api uint) error {
+	return setApi(api)
```
mheon: I don't see us calling this internally (and if we do, the public version has no locking). As such, I don't think we need the internal helper - you can just move the logic out into here.

tyhicks: Sure! That means that I'd have to move the cgo glue code to handle old versions of libseccomp into seccomp.go. The seccomp.go file doesn't have any cgo glue code like that currently which is why I opted to stick it into seccomp_internal.go. I'll move everything to seccomp.go.

tyhicks:
@mheon I've address all of your feedback. Here are the changes that I made:

* Move ActLog declaration below ActLog
* Point to seccomp_api_get(3) man page in GetApi() and SetApi() docs
* Provide more helpful error messages when GetLogBit() or SetLogBit() fail due to libseccomp being too old or the API level being set too low
* Update TestFilterAttributeGettersAndSetters() to call GetLogBit() and SetLogBit(), even when the API level is too low, and emit an "Ignoring failure" info message that includes the more helpful error messages mentioned above

mheon:
Code LGTM though I'd like to run some tests on my end to verify new functionality (and that we haven't broken earlier library versions). We're also still pending on final support landing in the library. Once that lands, ping me and I'll get this tested and merged.

tyhicks:
@mheon thanks! I'll describe the testing that I performed. I've tested with libseccomp-golang's `check` makefile target in the following scenarios:

1. New libseccomp-golang with old libseccomp and old kernel
1. New libseccomp-golang with new libseccomp and old kernel
1. New libseccomp-golang with new libseccomp and new kernel

When I say "new", I mean that the project has been updated with the respective logging patches and, in the case of userspace projects, API level patches. By "old", I mean that API level and logging patches are not applied.

All tests pass for each scenario.

tyhicks:
@mheon just an FYI that the libseccomp changes all are merged. No changes to my libseccomp-golang branch are needed so it is ready for your testing. Thanks!

mheon:
Alright, I'll try and get this merged sometime tomorrow then. Thanks!

tyhicks:
I've removed "[WIP]" from the title so that we don't forget that this PR is potentially ready to be merged.

mvo5:
I am keen to use this new logging feature in one of my projects. Anything I can help with to get this merged?

mheon:
Sorry for the delay here, it's been a busy few months. I'll get this tested and merged tonight

mheon:
Tests look good, but I noticed something concerning. Libseccomp just released v2.3.3, which is referenced a lot in this PR as the version that will include SCMP_ACT_LOG support (and API levels). Upstream v2.3.3 does not appear to have either of those, however.

@pcmoore What release are those likely to come out in? Can we throw v2.3.4 in there and be safe?

pcmoore:
@mheon the changes from Tyler are not simple bug fixes, but new functionality, so it would be the next "Y" release (given the X.Y.Z versioning).

Considering that the logging capability in the golang bindings needs to be conditional on the main libseccomp version, and we haven't released a libseccomp update with the logging functionality, it seems premature to add this to the golang bindings right now.

FWIW, I'm trying very hard to get a new libseccomp Y release ready, but I'm chasing some nasty bugs at the moment in the src/db.c code.  I'm getting very close, but not there yet.

mheon:
ACK. I'll hold off until you cut a new version with the appropriate defines, then.

tyhicks:
I found a bug in this PR that was introduced when I moved ActLog after ActAllow in the list of ScmpActions. I forgot to update actionEnd to be ActLog instead of ActAllow. This prevented ActLog from being used as the default action for a filter. I've incorporated that change into the `Add support for SCMP_ACT_LOG` commit and updated this PR.

blacktop:
Is there any reason this isn't able to be merged in now?

blacktop:
Looks like you guys have a 2.4.0 release now 😁 @mheon 🙏

blacktop:
ping?

pcmoore:
@mheon are you waiting on anything else from the main libseccomp library?

seccomp_internal.go
```diff
@@ -122,6 +137,26 @@ unsigned int get_micro_version()
 }
 #endif
 
+// The libseccomp API level functions were added in v2.3.3
+#if (SCMP_VER_MAJOR < 2) || \
```
mheon: Should we swap this macro to require 2.4.0 minimum?

mheon:
Nope, I think this one is set, now that 2.4.0 is released.

Sorry for the delay here, been busy with other projects. I have one lingering issue with the version check macros, but otherwise this seems set.

@tyhicks If possible, can you push a new version with the version check macros set for 2.4.0 instead? If not, I can handle the changes when I merge.

seccomp_internal.go
tyhicks: Nice catch!

tyhicks:
> @tyhicks If possible, can you push a new version with the version check macros set for 2.4.0 instead? If not, I can handle the changes when I merge.

@mheon I can do that. Is a force push to the same branch alright with you?

mheon:
Works fine for me!

tyhicks:
@mheon Done! Note that I didn't retest any of it since I no longer have the environment set up to do so. These were all simple text changes and the `master` branch hasn't changed since I last tested and pushed this PR so I think it'll be fine.

seccomp_internal.go
tyhicks: There were two macros to swap and several comments that needed to be adjusted from `v2.3.3` to `v2.4.0`.

mheon:
I'll run the test suite locally to verify and push the commit to master once that's done

mheon:
Had some local testing hiccups, but it looks like it was purely environment - the tests were all passing, but failing to write results. Took a bit to sort out, but the tests are passing. Pushed to master. Thanks for your patience here!

PR #41, opened 2020-03-13
RFE: initial Travis CI support

pcmoore:
This is basic, but it gets us started.


pcmoore:
I went ahead and merged this at HEAD 4387a10873b9fbdef54045302e182e76e1105d76 so we would have something in place, we can improve upon in later.

PR #50, opened 2020-07-06
RFE: seccomp userspace notification

rodnymolina:
These changes require the latest [seccomp.h](https://github.com/seccomp/libseccomp/blob/master/include/seccomp.h.in) header from libseccomp's [master](https://github.com/seccomp/libseccomp) branch for proper compilation.

These commits add the seccomp userspace notification API present in version >= 2.4 of the libseccomp library.

This API allows userspace to get a notification when a filter configured with a notification action triggers. The trigger suspends processing of the syscall until the notification is delivered to userspace and acknowledged back.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -2,6 +2,9 @@ libseccomp-golang: Releases
 ===============================================================================
 https://github.com/seccomp/libseccomp-golang
 
+* Version 1.0.0 - Dec 12, 2019
+- Added support for the seccomp userspace notification API.
+
 * Version 0.9.1 - May 21, 2019
 - Minimum supported version of libseccomp bumped to v2.2.0
 - Use Libseccomp's `seccomp_version` API to retrieve library version
--- a/seccomp.go
+++ b/seccomp.go
@@ -239,6 +239,9 @@ const (
 
 const (
 	// Userspace notification response flags
+
+	// Tells the kernel to continue executing the system call that triggered the
+	// notification. Must only be used when the notication response's error is 0.
 	NotifRespFlagContinue uint32 = 1
 )
 
--- a/seccomp.go
+++ b/seccomp.go
@@ -20,9 +20,10 @@ import (
 
 // C wrapping code
 
-// #cgo pkg-config: libseccomp
+// #cgo CFLAGS: -I../seccomp/include
+// #cgo LDFLAGS: -L../seccomp/src/.libs -lseccomp
 // #include <stdlib.h>
-// #include <seccomp.h>
+// #include "seccomp.h"
 import "C"
 
 // Exported types
@@ -69,8 +70,39 @@ type ScmpCondition struct {
 	Operand2 uint64        `json:"operand_two,omitempty"`
 }
 
-// ScmpSyscall represents a Linux System Call
+// Seccomp userspace notification structures associated with filters that use the ActNotify action.
+
+// ScmpSyscall identifies a Linux System Call by its number.
 type ScmpSyscall int32
+
+// ScmpFd represents a file-descriptor used for seccomp userspace notifications.
+type ScmpFd int32
+
+// ScmpNotifData describes the system call context that triggered a notification.
+type ScmpNotifData struct {
+	Syscall      ScmpSyscall `json:"syscall,omitempty"`
+	Arch         ScmpArch    `json:"arch,omitempty"`
+	InstrPointer uint64      `json:"instr_pointer,omitempty"`
+	Args         []uint64    `json:"args,omitempty"`
+}
+
+// ScmpNotifReq represents a seccomp userspace notification. See NotifReceive() for
+// further details.
+type ScmpNotifReq struct {
+	Id    uint64        `json:"id,omitempty"`
+	Pid   uint32        `json:"pid,omitempty"`
+	Flags uint32        `json:"flags,omitempty"`
+	Data  ScmpNotifData `json:"data,omitempty"`
+}
+
+// ScmpNotifResp represents a seccomp userspace notification response. See NotifRespond()
+// for further details.
+type ScmpNotifResp struct {
+	Id    uint64 `json:"id,omitempty"`
+	Val   uint64 `json:"val,omitempty"`
+	Error int32  `json:"error,omitempty"`
+	Flags uint32 `json:"flags,omitempty"`
+}
 
 // Exported Constants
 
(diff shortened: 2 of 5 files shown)


pcmoore:
Hi @rodnymolina, thanks for the PR!  @drakenclimber and myself are busy right now working on the main libseccomp v2.5.0 release, but once that is done we will be turning our focus to the outstanding libseccomp-golang issues.

rhatdan:
@vrothberg @giuseppe PTAL

CHANGELOG
```diff
@@ -2,6 +2,9 @@ libseccomp-golang: Releases
 ===============================================================================
 https://github.com/seccomp/libseccomp-golang
 
+* Version 1.0.0 - Dec 12, 2019
```
giuseppe, line 5: is this correct?

seccomp_internal.go
```diff
@@ -14,7 +14,8 @@ import (
 // Get the seccomp header in scope
 // Need stdlib.h for free() on cstrings
 
-// #cgo pkg-config: libseccomp
+// #cgo CFLAGS: -I../seccomp/include
+// #cgo LDFLAGS: -L../seccomp/src/.libs -l:libseccomp.a
```
giuseppe, line 18: same here, let's use pkg-config

seccomp.go
```diff
@@ -20,7 +20,8 @@ import (
 
 // C wrapping code
 
-// #cgo pkg-config: libseccomp
+// #cgo CFLAGS: -I../seccomp/include
+// #cgo LDFLAGS: -L../seccomp/src/.libs -l:libseccomp.a
```
giuseppe, line 24: I think we should not be dropping pkg-config here

seccomp.go
```diff
@@ -947,3 +1052,100 @@ func (f *ScmpFilter) ExportBPF(file *os.File) error {
...
+	}
+
+	// we only use the request here; the response is unused
+	if retCode := C.seccomp_notify_alloc(&req, &resp); retCode != 0 {
+		return nil, errRc(retCode)
+	}
+
+	defer func() {
+		C.free(unsafe.Pointer(req))
+		C.free(unsafe.Pointer(resp))
```
giuseppe, line 1100: I think we should be using `seccomp_notify_free` here instead of plain `free`

seccomp.go
```diff
@@ -947,3 +1052,100 @@ func (f *ScmpFilter) ExportBPF(file *os.File) error {
...
+	}
+
+	// we only use the reponse here; the request is discarded
+	if retCode := C.seccomp_notify_alloc(&req, &resp); retCode != 0 {
+		return errRc(retCode)
+	}
+
+	defer func() {
+		C.free(unsafe.Pointer(req))
+		C.free(unsafe.Pointer(resp))
```
giuseppe, line 1127: same here, let's use `seccomp_notify_free`

seccomp_internal.go
```diff
@@ -88,6 +89,7 @@ const uint32_t C_ACT_ERRNO         = SCMP_ACT_ERRNO(0);
 const uint32_t C_ACT_TRACE         = SCMP_ACT_TRACE(0);
 const uint32_t C_ACT_LOG           = SCMP_ACT_LOG;
 const uint32_t C_ACT_ALLOW         = SCMP_ACT_ALLOW;
+const uint32_t C_ACT_NOTIFY        = SCMP_ACT_NOTIFY;
```
alban, line 92: To follow the code pattern of lines 72-82, you could add the following before:
```
#ifndef SCMP_ACT_USER_NOTIF
#define SCMP_ACT_USER_NOTIF 0x7fc00000U
#endif
```

seccomp_test.go
```diff
@@ -701,3 +677,189 @@ func TestCreateActKillProcessFilter(t *testing.T) {
...
+	}
+
+	filter, err := NewFilter(ActAllow)
+	if err != nil {
+		t.Errorf("Error creating filter: %s", err)
+	}
+	defer filter.Release()
+
+	// Seccomp notification is only supported on single-thread filters
+	err = filter.SetTsync(false)
```
alban, line 798: Linux v5.7 adds support for TSYNC and USER_NOTIF together:
https://github.com/torvalds/linux/commit/51891498f2da78ee64dfad88fa53c9e85fb50abf

It would be nice to test that.

seccomp.go
```diff
@@ -540,6 +626,27 @@ func NewFilter(defaultAction ScmpAction) (*ScmpFilter, error) {
...
+func (f *ScmpFilter) SetTsync(val bool) error {
+	var cval C.uint32_t
+
+	if val == true {
+		cval = 1
+	} else {
+		cval = 0
+	}
+
+	err := f.setFilterAttr(filterAttrTsync, cval)
```
alban, line 643: From seccomp api level 6, we can use `SECCOMP_FILTER_FLAG_TSYNC_ESRCH` to allow tsync and notify together. I think `SetTsync()` should check `apiLevel`, and if appropriate use `filterAttrTsync` and `filterAttrTsyncEsrch` together.

Linux v5.7 adds support for TSYNC and USER_NOTIF together:
https://github.com/torvalds/linux/commit/51891498f2da78ee64dfad88fa53c9e85fb50abf

seccomp_test.go
```diff
@@ -525,6 +530,7 @@ func TestMergeFilters(t *testing.T) {
 }
 
 func TestRuleAddAndLoad(t *testing.T) {
+
```
alban, line 533: unneeded empty line (and in other functions in this file below)

seccomp_test.go
```diff
@@ -701,3 +677,189 @@ func TestCreateActKillProcessFilter(t *testing.T) {
...
+		t.Errorf("Error loading filter: %s", err)
+	}
+
+	fd, err := filter.GetNotifFd()
+	if err != nil {
+		t.Errorf("Error getting filter notification fd: %s", err)
+	}
+
+	if fd < 3 {
+		t.Errorf("Error in notification fd: want 0, got %v", fd)
```
alban, line 829:
```suggestion
		t.Errorf("Error in notification fd: want >=3, got %v", fd)
```

alban (approved):
Some comments but LGTM

alban:
Travis fails because this it is tested against an old version of libseccomp. It needs #58 to run on libseccomp-2.5.0.

I fixed some other test failures in my branch [alban/userspace_notification](https://github.com/kinvolk/libseccomp-golang/commits/alban/userspace_notification).

seccomp_internal.go
alban: Agree.

It's possible to test `libseccomp-golang` against `libseccomp-2.5.0`, even if it is not installed, with the following:

```
cd ../libseccomp && mkdir -p prefix
./configure --prefix=$PWD/prefix && make && make install
cd ../libseccomp-golang
PKG_CONFIG_PATH=$PWD/../libseccomp/prefix/lib/pkgconfig/ make
LD_PRELOAD=$PWD/../libseccomp/prefix/lib/libseccomp.so.2.5.0 PKG_CONFIG_PATH=$PWD/../libseccomp/prefix/lib/pkgconfig/ make test
```

alban:
I added my patches in #59.

pcmoore:
Considering PR #59 I think it makes sense to close out this PR, if anyone has any objections please shout and we can reopen it - thanks!

PR #51, opened 2020-07-22
RFE: add support for SCMP_FLTATR_CTL_SSB

KentaTada:
Create a new scmpFilterAttr, filterAttrSSB, to represent libseccomp's
SCMP_FLTATR_CTL_SSB

This filter is only supported in libseccomp 2.5.0 and newer
with API level 4 or higher.

Signed-off-by: Kenta Tada <Kenta.Tada@sony.com>

### DIFF
--- a/seccomp.go
+++ b/seccomp.go
@@ -793,6 +793,30 @@ func (f *ScmpFilter) GetLogBit() (bool, error) {
 	return true, nil
 }
 
+// GetSSB returns the current state the SSB bit will be set to on the filter
+// being loaded, or an error if an issue was encountered retrieving the value.
+// The SSB bit tells the kernel that a seccomp user is not interested in enabling
+// Speculative Store Bypass mitigation.
+// The SSB bit is only usable when libseccomp API level 4 or higher is
+// supported.
+func (f *ScmpFilter) GetSSB() (bool, error) {
+	ssb, err := f.getFilterAttr(filterAttrSSB)
+	if err != nil {
+		api, apiErr := getAPI()
+		if (apiErr != nil && api == 0) || (apiErr == nil && api < 4) {
+			return false, fmt.Errorf("getting the SSB flag is only supported in libseccomp 2.5.0 and newer with API level 4 or higher")
+		}
+
+		return false, err
+	}
+
+	if ssb == 0 {
+		return false, nil
+	}
+
+	return true, nil
+}
+
 // SetBadArchAction sets the default action taken on a syscall for an
 // architecture not in the filter, or an error if an issue was encountered
 // setting the value.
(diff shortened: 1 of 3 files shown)


rhatdan:
@giuseppe PTAL

seccomp.go
```diff
@@ -793,6 +793,30 @@ func (f *ScmpFilter) GetLogBit() (bool, error) {
 	return true, nil
 }
 
+// GetSsbBit returns the current state the SSB bit will be set to on the filter
+// being loaded, or an error if an issue was encountered retrieving the value.
+// The SSB bit tells the kernel that a seccomp user is not interested in enabling
+// Speculative Store Bypass mitigation.
+// The Log bit is only usable when libseccomp API level 4 or higher is
+// supported.
+func (f *ScmpFilter) GetSsbBit() (bool, error) {
```
pcmoore, line 802: Can we just call this `GetSSB()`?  I dislike the forced lowercasing of acronyms in strict camel case.

seccomp.go
```diff
@@ -841,6 +865,28 @@ func (f *ScmpFilter) SetLogBit(state bool) error {
 	return err
 }
 
+// SetSsbBit sets the state of the SSB bit, which will be applied on filter
+// load, or an error if an issue was encountered setting the value.
+// The Log bit is only usable when libseccomp API level 4 or higher is
+// supported.
+func (f *ScmpFilter) SetSsbBit(state bool) error {
```
pcmoore, line 872: Similarly, let's call this `SetSSB()`.

seccomp.go
```diff
@@ -793,6 +793,30 @@ func (f *ScmpFilter) GetLogBit() (bool, error) {
...
+// The SSB bit tells the kernel that a seccomp user is not interested in enabling
+// Speculative Store Bypass mitigation.
+// The Log bit is only usable when libseccomp API level 4 or higher is
+// supported.
+func (f *ScmpFilter) GetSsbBit() (bool, error) {
+	ssb, err := f.getFilterAttr(filterAttrSSB)
+	if err != nil {
+		api, apiErr := getAPI()
+		if (apiErr != nil && api == 0) || (apiErr == nil && api < 4) {
+			return false, fmt.Errorf("getting the log bit is only supported in libseccomp 2.5.0 and newer with API level 4 or higher")
```
pcmoore, line 807: You don't mean to say "log bit" here do you?  You mean "SSB flag", right?

seccomp.go
```diff
@@ -841,6 +865,28 @@ func (f *ScmpFilter) SetLogBit(state bool) error {
...
+
+	if state {
+		toSet = 0x1
+	}
+
+	err := f.setFilterAttr(filterAttrSSB, toSet)
+	if err != nil {
+		api, apiErr := getAPI()
+		if (apiErr != nil && api == 0) || (apiErr == nil && api < 4) {
+			return fmt.Errorf("setting the log bit is only supported in libseccomp 2.5.0 and newer with API level 4 or higher")
```
pcmoore, line 883: Similar to above.

seccomp.go
KentaTada: Changed

seccomp.go
KentaTada: Changed

seccomp.go
KentaTada: I'm sorry. I fixed it.

seccomp.go
KentaTada: Fixed

pcmoore:
Thanks @KentaTada.

@drakenclimber thoughts?

pcmoore:
@KentaTada please look into and resolve the Travis CI failure.

KentaTada:
> @KentaTada please look into and resolve the Travis CI failure.

Maybe, I think that the test environment needs libseccomp v2.5.0 or later.
I will investigate more deeply next week.

pcmoore:
> Maybe, I think that the test environment needs libseccomp v2.5.0 or later.

Ah, yes, that is probably it.  At the very least please make sure that the test checks the installed libseccomp version and adjusts accordingly (e.g. skips tests which require newer functionality that is not supported).  Part of PR #59 brings in newer versions of libseccomp into the test but unfortunately it is wrapped up in that monster PR; I'll look into merging just that portion of the PR as it will likely help you as well.

KentaTada:
I skip this test when the API level is less than 4.
But I want to modify the whole of TestFilterAttributeGettersAndSetters() because I want to modify the implementation of the API check for SetLogBit() and GetLogBit() like this commit and reduce the duplicated GetAPI().
This PR is to add the support for SCMP_FLTATR_CTL_SSB. So, I may refactor the test code after this commit is merged.

drakenclimber (approved):
Looks good to me.  Thanks!

pcmoore:
Merged via 276b0ea443de00461570ec512734ec0c3bc6cc42, thanks for your patience @KentaTada!

Issue #9, opened 2016-05-25
RFE: add support for parisc/parisc64


pcmoore:
We have parisc[64] support in the main libseccomp repository and it is scheduled for the v2.4 release.

pcmoore:
Resolved via PR #53.

PR #53, opened 2020-07-23
RFE: add defines for PARISC / PARISC64 goloang bindings

KentaTada:
I want to close https://github.com/seccomp/libseccomp-golang/issues/9 to release v0.9.2

Signed-off-by: Kenta Tada <Kenta.Tada@sony.com>

### DIFF
--- a/seccomp.go
+++ b/seccomp.go
@@ -117,6 +117,10 @@ const (
 	ArchS390 ScmpArch = iota
 	// ArchS390X represents 64-bit System z/390 syscalls
 	ArchS390X ScmpArch = iota
+	// ArchPARISC represents 32-bit PA-RISC
+	ArchPARISC ScmpArch = iota
+	// ArchPARISC64 represents 64-bit PA-RISC
+	ArchPARISC64 ScmpArch = iota
 )
 
 const (
@@ -223,6 +227,10 @@ func GetArchFromString(arch string) (ScmpArch, error) {
 		return ArchS390, nil
 	case "s390x":
 		return ArchS390X, nil
+	case "parisc":
+		return ArchPARISC, nil
+	case "parisc64":
+		return ArchPARISC64, nil
 	default:
 		return ArchInvalid, fmt.Errorf("cannot convert unrecognized string %q", arch)
 	}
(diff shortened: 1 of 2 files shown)


pcmoore (approved):

pcmoore:
Still not a golang expert yet, but this looks reasonable to me.  Any objections @drakenclimber?

seccomp.go
```diff
@@ -117,6 +117,10 @@ const (
 	ArchS390 ScmpArch = iota
 	// ArchS390X represents 64-bit System z/390 syscalls
 	ArchS390X ScmpArch = iota
+	// ArchPARISC represents 32-bit PA-RISC
+	ArchPARISC ScmpArch = iota
```
drakenclimber, line 121: Foolish question from a golang neophyte - would we be better off setting the first value (ArchInvalid in this case) equal to `iota` and leaving the rest blank.

seccomp_internal.go
```diff
@@ -210,7 +220,7 @@ const (
 	scmpError C.int = -1
 	// Comparison boundaries to check for architecture validity
 	archStart ScmpArch = ArchNative
-	archEnd   ScmpArch = ArchS390X
+	archEnd   ScmpArch = ArchPARISC64
```
drakenclimber, line 223: Another newbie question here - this feels error prone.  In C, we would often use a value at the end of the enumeration (say ARCH_LAST or something) so that the actual code doesn't need to change.  What is golang's stance on this?

drakenclimber (commented):
Just a couple foolish questions from me.

drakenclimber:
> Just a couple foolish questions from me.

And if they happen to be valid concerns (gasp), they obviously don't need to be addressed as part of this parisc review.

drakenclimber (approved):

seccomp.go
pcmoore: I *thought* the convention was to use `iota` for all the entries, and the current use seems to bear that out.  However, like you I am far from a golang expert.

seccomp_internal.go
pcmoore: I'm not sure how most golang folks do this, but it certainly fits with how we've handled it thus far so I see no reason to hold this PR up.  However, I do agree that it seems fragile.

pcmoore:
Merged via 4b7f544e4bee8c201c9d76dcf7ef8d8e96afbfea, thanks @KentaTada!

PR #57, opened 2020-08-24
RFE: add libseccomp notifier API

### LINKED ISSUE #56: RFE: add seccomp notifier support
I saw that libseccomp 2.5.0 added 6 new notifier APIs:
```c
int seccomp_notify_alloc(struct seccomp_notif **req, struct seccomp_notif_resp **resp);
void seccomp_notify_free(struct seccomp_notif *req, struct seccomp_notif_resp *resp);
int seccomp_notify_receive(int fd, struct seccomp_notif *req);
int seccomp_notify_respond(int fd, struct seccomp_notif_resp *resp);
int seccomp_notify_id_valid(int fd, uint64_t id);
int seccomp_notify_fd(const scmp_filter_ctx ctx);
```
Referring to https://github.com/seccomp/libseccomp-golang/issues/37, I think it would make sense to add the binding and structures to this project, too. I could take care of the implementation/testing if you want. WDYT?

saschagrunert:
Just drafting this PR to propose an API surface. If we agree on the API, then I'll continue with testing. :)
Fixes #56

### DIFF
--- a/notify.go
+++ b/notify.go
@@ -1,0 +1,69 @@
+// +build linux
+
+package seccomp
+
+import "os"
+
+// #cgo pkg-config: libseccomp
+// #include <stdlib.h>
+// #include <seccomp.h>
+import "C"
+
+// ScmpNotifReq is a seccomp notifier request
+type ScmpNotifReq struct {
+	inner C.seccomp_notif
+}
+
+// ScmpNotifResp is a seccomp notifier response
+type ScmpNotifResp struct {
+	inner C.seccomp_notif_resp
+}
+
+// NotifyAlloc creates a new pair of notification request/response structures.
+// See libseccomp API `seccomp_notify_alloc()`.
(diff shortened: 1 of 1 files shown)


notify.go
```diff
@@ -0,0 +1,69 @@
...
+// Receive tries to get a notification from a seccomp notification file.
+// See libseccomp API `seccomp_notify_receive()`.
+func (s *ScmpNotifReq) Receive(file *os.File) error {
+	if ret := C.seccomp_notify_receive(file.Fd(), s.inner); ret != 0 {
+		return errRc(ret)
+	}
+	return nil
+}
+
+// Repond tries to send a notification to a seccomp notification file.
```
drakenclimber, line 48: `make lint` is failing because of a typo here

`s/Repond/Respond`

saschagrunert: Thanks for the hint! :pray: Fixed that issue.

alban:
https://github.com/seccomp/libseccomp-golang/pull/50 is covering the notifier API and has some tests.

PR #58, opened 2020-09-26
Travis: compile with libseccomp-2.5.0

alban:
Run the tests with libseccomp-2.5.0 instead of the version in Ubuntu Bionic. This is needed to test #50.


alban:
A new version of this patch is included in #59

PR #59, opened 2020-09-28
RFE: seccomp userspace notification (follow-up to #50)

alban:
This PR is based on @rodnymolina's work from #50 and adds the following:
- #58 to update Travis tests to libseccomp 2.5.0
- Address reviews
- Fix tests

cc @giuseppe


giuseppe:
thanks for addressing all the comments.  LGTM

giuseppe (approved):

.travis.yml
```diff
@@ -23,15 +23,20 @@ addons:
...
 script:
+  - wget https://github.com/seccomp/libseccomp/releases/download/v2.5.0/libseccomp-2.5.0.tar.gz
+  - echo 463b688bf7d227325b5a465b6bdc3ec4 libseccomp-2.5.0.tar.gz | md5sum -c
+  - tar xf libseccomp-2.5.0.tar.gz
+  - pushd libseccomp-2.5.0 && ./configure && make && sudo make install && popd
   - make check-syntax
   - make lint
-  - make check
+  - make vet
+  - LD_PRELOAD=/usr/local/lib/libseccomp.so.2.5.0 make test
```
AkihiroSuda, line 42: Could this be LD_LIBRARY_PATH without preloading?

alban: Yes it can. I updated the branch to install in a prefix, and use PKG_CONFIG_PATH and LD_LIBRARY_PATH (without LD_PRELOAD).

AkihiroSuda (approved):

alban:
@pcmoore @drakenclimber should the commits be squashed into logical commits or would you do squash in a single commit when merging?

@rodnymolina @ctalledo some of the initial commits from #50 don't have the "Signed-off-by" line. Are you ok with adding it?

rodnymolina:
@alban, thanks for picking this one up and addressing all the review comments, haven't had cycles to do it myself. Do you want us to add the signed-off to our original commits and re-submit them again? Or there's another (quicker) way to do this?

alban:
I could just edit your commits in this PR with the following if you give your ok
```
Signed-off-by: Rodny Molina <rmolina@nestybox.com>
```

rodnymolina:
Oh sure, please go ahead. Thanks!

alban:
I forced-pushed the branch with the following changes:
- Add Signed-off-by tag on commitmsgs in agreement with Rodny and Cesar
- Add compatibility with old libseccomp (<2.5.0). This is done using the existing mechanism with `#if SCMP_VER_*` in seccomp_internal.go. In order to keep seccomp.go simple without additional cgo code, the new public API about notify added in seccomp.go is now a simple wrapper to internal functions in seccomp_internal.go.
- Travis CI: add a matrix to check that it works with 3 different versions of seccomp, back to 2.2.1
- Unit tests: skip TestNotif on old seccomp and add new test TestNotifUnsupported to check that error management is done correctly on old seccomp.
- Fix spelling of SCMP_ACT_NOTIFY (remove SCMP_ACT_USER_NOTIF)

With the changes above, I am able to use this libseccomp-golang branch successfully in runc (tested with Travis in CentOS 7, Ubuntu Bionic, Fedora 32).

seccomp_test.go
```diff
@@ -701,3 +675,265 @@ func TestCreateActKillProcessFilter(t *testing.T) {
...
+	seccompFdChan := make(chan ScmpFd)
+	errorChan := make(chan error, 2)
+	infoChan := make(chan string)
+	done := make(chan struct{})
+
+	go func() {
+		// Lock this goroutine to it's current kernel thread; otherwise the go runtime may
+		// switch us to a different OS thread, bypassing the seccomp notification filter.
+		runtime.LockOSThread()
+		defer runtime.UnlockOSThread()
```
yvesf, line 860: Documentation of `UnlockOSThread` says:
> // Before calling UnlockOSThread, the caller must ensure that the OS
// thread is suitable for running other goroutines. If the caller made
// any permanent changes to the state of the thread that would affect
// other goroutines, it should not call this function and thus leave
// the goroutine locked to the OS thread until the goroutine (and
// hence the thread) exits.

Since we don't _unload_ this filter (because it's not possible), I assume it would be correct to not call UnlockOSThread() here. Otherwise it could happen that by random choice this thread is re-used. In this test this seems improable but if it's true then it would be good to give the right example here. I'm not an expert and just reasoning from the docs, I have no proof of this.

yvesf:
Maybe someone else could verify if this affects only me:
When I run the test for notifications more than once: `go test -v . -count 2 -v -run 'TestNotif$'`
... it fails starting from the second run. Almost like the filter is not loaded. The notify-fd even returns "bad filedescriptor" when I try to close it. From the code it's not clear to me why it could fail silently like this.
 (running debian buster with seccomp libs 2.5.0-1 from experimental in docker within kernel 5.8.13-zen1 on nixos 20.09).

seccomp_test.go
ctalledo: Thanks, that's a good catch. I had missed that description on UnlockOSThread(), and I agree with your reasoning that we should not unlock the goroutine from the OS thread given that the seccomp filter is already acting on it.

alban: @yvesf Thanks for testing. I have the same error when trying `-count 2` like you did. I tried to remove the call to `UnlockOSThread` but it didn't help.

I observed the following with the command:
```
strace -f -s 256 -e chdir,seccomp go test -v . -count 2 -v -run 'TestNotif$'
```

* First iteration of the test:
```
[pid 1901155] seccomp(SECCOMP_SET_MODE_FILTER, SECCOMP_FILTER_FLAG_NEW_LISTENER, {len=9, filter=0x1a70580}) = 6
[pid 1901155] chdir("/non-existent-path"    <unfinished ...>
[pid 1901155] <... chdir resumed>)      = -1 ENOENT (No such file or directory)
```
-> seccomp() returns the file descriptor correctly
-> chdir() returned ENOENT as expected

* Second iteration of the test:
```
[pid 1901157] seccomp(SECCOMP_SET_MODE_FILTER, 0, {len=9, filter=0x7f6948001b50}) = 0
[pid 1901157] chdir("/non-existent-path") = -1 ENOSYS (Function not implemented)
```
-> The pid is different, so removing `UnlockOSThread` worked
-> seccomp() is not passed the flag SECCOMP_FILTER_FLAG_NEW_LISTENER as expected (why?)
-> seccomp() fails to return the file descriptor
-> chdir() does not return ENOENT but ENOSYS.

I am continuing to investigate...

alban:
@yvesf I found the root cause: this is because `sys_filter_load()` ([see code](https://github.com/seccomp/libseccomp/blob/02812f99e8d1df2e671dac675b4af663d0266303/src/system.c#L375-L376)) uses global variable (`state.notify_fd`) to remember if a seccomp fd has already been set. There is a function exposed in the API to reset that global variable:

From seccomp_reset(3):

> If seccomp_reset() is called with a NULL filter, it resets the library's global task state, including any notification file descriptors
retrieved by seccomp_notify_fd(3).  Normally this is not needed, but it may be required to continue using the library after a fork() or clone() call to ensure the API level and user notification state is properly reset.

 I could fix the issue in libseccomp-golang with the following patch:

```diff
diff --git a/seccomp.go b/seccomp.go
index 2c12662..68e7690 100644
--- a/seccomp.go
+++ b/seccomp.go
@@ -682,6 +682,14 @@ func (f *ScmpFilter) Reset(defaultAction ScmpAction) error {
        return nil
 }
 
+func GlobalReset() error {
+       if retCode := C.seccomp_reset(nil, 0); retCode != 0 {
+               return errRc(retCode)
+       }
+
+       return nil
+}
+
 // Release releases a filter context, freeing its memory. Should be called after
 // loading into the kernel, when the filter is no longer needed.
 // After calling this function, the given filter is no longer valid and cannot
diff --git a/seccomp_test.go b/seccomp_test.go
index 2b82835..5ca9781 100644
--- a/seccomp_test.go
+++ b/seccomp_test.go
@@ -743,6 +743,8 @@ func notifHandler(ch chan error, fd ScmpFd, tests []notifTest) {
 }
 
 func TestNotif(t *testing.T) {
+       GlobalReset()
+
        // seccomp notification requires API level >= 5
        api, err := GetAPI()
        if err != nil {
@@ -857,7 +859,9 @@ func TestNotif(t *testing.T) {
                // Lock this goroutine to it's current kernel thread; otherwise the go runtime may
                // switch us to a different OS thread, bypassing the seccomp notification filter.
                runtime.LockOSThread()
-               defer runtime.UnlockOSThread()
+
+               // Don't call runtime.UnlockOSThread() to ensure that the thread with the
+               // seccomp filter applied will not be reused in another test.
 
                err = filter.Load()
                if err != nil {
```

I think libseccomp API is not ideal with this global state. Although it can make sense in single-threaded programs, it could cause difficulties if a multi-threaded program ever wants to setup different filters for different threads. I would like if libseccomp API could be improved, but in order for libseccomp-golang to support libseccomp-2.5.0, I would add this GlobalReset() in the Go API. What do you think?

alban:
Even with the fix from above, we can't run some tests with `filter.SetTsync(false)` (`TestNotif`) and some tests with `filter.SetTsync(true)` (`TestLogAct`) because once the threads are not using the same seccomp filters anymore, the kernel will refuse to use `SECCOMP_FILTER_FLAG_TSYNC`:

> If  any  thread cannot synchronize to the same filter tree, the call will not attach the new seccomp filter, and will fail

This can reproduce with:
```
go test -v . -count 2 -v -run 'TestLogAct|TestNotif$'
```

For this, the tests using tsync and the tests not using tsync should be executed separately (in different processes).

yvesf:
> Even with the fix from above, we can't run some tests with `filter.SetTsync(false)` (`TestNotif`) and some tests with `filter.SetTsync(true)` (`TestLogAct`) because once the threads are not using the same seccomp filters anymore
...
> For this, the tests using tsync and the tests not using tsync should be executed separately (in different processes).

Not sure if it's too much of a hack but we could execute the test in subprocesses, go's `syscall/syscall_linux_test.go` does something similar.
```go
func TestLogAct(t *testing.T) {
	execInSubprocess(t, func(t *testing.T) {
		expectedPid := syscall.Getpid()
```

<details>

<summary>execInSubprocess(t, func)</summary>

```go
// execInSubprocess calls the go test binary again for the same test.
// This must be only top-level statment in the test function. Do not nest this.
// It will slightly defect the test log output as the test is entered twice
func execInSubprocess(t *testing.T, f func(t *testing.T)) {
	const subprocessEnvKey = `GO_SUBPROCESS_KEY`
	if testIDString, ok := os.LookupEnv(subprocessEnvKey); ok && testIDString == "1" {
		t.Run(`subprocess`, f)
		return
	}

	cmd := exec.Command(os.Args[0])
	cmd.Args = []string{os.Args[0], "-test.run=" + t.Name() + "$", "-test.v=true"}
	for _, arg := range os.Args {
		if strings.HasPrefix(arg, `-test.testlogfile=`) {
			cmd.Args = append(cmd.Args, arg)
		}
	}
	cmd.Env = []string{subprocessEnvKey + "=1"}
	cmd.Stdin = os.Stdin
	cmd.Stdout = os.Stdout
	cmd.Stderr = os.Stderr

	err := cmd.Start()
	if err != nil {
		t.Error("failed to spawn test in sub-process", err)
		t.FailNow()
	}

	err = cmd.Wait()
	if err != nil {
		if err, ok := err.(*exec.ExitError); ok {
			t.Errorf("Test failed with status %d", err.ExitCode())
		}
		t.Error(`test failed`)
		t.FailNow()
	}
}
```

</details>

alban:
@yvesf Thanks, it works for me! I pushed a new commit using your `execInSubprocess` function. I changed stdout / stderr so that the recursive output is aligned more nicely. I didn't find the same code pattern in `syscall/syscall_linux_test.go` though. Do I reference the code correctly in the commitmsg?

yvesf:
> I didn't find the same code pattern in `syscall/syscall_linux_test.go` though

It's not the same pattern, just similar. TestLinuxDeathSignal for instance calls itself with `GO_DEATHSIG_PARENT` env variable set. In the `TestMain` this env variable is detected and execution redirected to `deathSignalChild()`.

seccomp_test.go
```diff
@@ -5,11 +5,63 @@
...
+	cmd.Args = []string{os.Args[0], "-test.run=" + t.Name() + "$", "-test.v=true"}
+	for _, arg := range os.Args {
+		if strings.HasPrefix(arg, `-test.testlogfile=`) {
+			cmd.Args = append(cmd.Args, arg)
+		}
+	}
+	cmd.Env = []string{subprocessEnvKey + "=1"}
+	cmd.Stdin = os.Stdin
+
+	var b bytes.Buffer
```
yvesf, line 40: using `strings.Builder` instead of `bytes.Buffer` would not require the new import and not the `string(b.Bytes())` cast.

alban: Updated.

alban:
Gentle reminder about this PR. Thanks!

pcmoore:
Wow, 27 patches ... that's something.  Especially considering I think you could count the associated main library patches to do enable userspace notification on one hand :)

I'm not sure yet if it is the case with this PR, but for future submissions please take the approach of simply updating flawed patches in the PR if the PR is still open and not merged.  I generally dislike merging commits which are known to be broken; it both makes bisection difficult and the git log overly noisy.

All that said, I'll start taking a look at this now, but given the number of patches it make take some time.

CHANGELOG
```diff
@@ -2,6 +2,9 @@ libseccomp-golang: Releases
 ===============================================================================
 https://github.com/seccomp/libseccomp-golang
 
+* Version UNRELEASED - FUTURE
```
pcmoore, line 5: I appreciate the gesture, but please refrain from adding things to the CHANGELOG, @drakenclimber and I update that as part of the release process.

Makefile
```diff
@@ -19,7 +19,7 @@ vet:
 	go vet -v
 
 test:
-	go test -v
+	go test -v -timeout 10s -count 2
```
pcmoore, line 22: Oooh, *magic* numbers, fun!  With stuff like this I think it is a good idea to briefly mention in the code why a timeout of 10s is necessary and why you picked a count of two.

seccomp.go
```diff
@@ -34,19 +41,27 @@ type VersionError struct {
 	minimum string
 }
 
+// Caches the libseccomp API level
+var apiLevel uint
+
+func init() {
+	// This forces the cgo libseccomp to initialize its internal API support state,
+	// which is necessary in order to other seccomp APIs to work correctly.
+	GetAPI()
```
pcmoore, line 50: That isn't the intention and I *think* we resolved this in the latest v2.5.x release, however I will concede that this is likely necessary for older versions of libseccomp.  Perhaps update the comment to mention this is necessary on *older* versions of libseccomp?

CHANGELOG
alban: ok, I will remove that change.

Makefile
alban: Makes sense. I'll add the following explanation:
```diff
+# Previous bugs have made the tests freeze until the timeout. Golang default
+# timeout for tests is 10 minutes, which is too long, considering current tests
+# can be executed in less than 1 second. Reduce the timeout, so problems can
+# be noticed earlier in the CI.
+TEST_TIMEOUT=10s
+
+# Some tests run with SetTsync(false) and some tests with SetTsync(true). Once
+# the threads are not using the same seccomp filters anymore, the kernel will
+# refuse to use Tsync, causing next tests to fail. This issue could be left
+# unnoticed if the test with SetTsync(false) is executed last.
+#
+# Run tests twice ensure that no test leave the testing process in a state
+# unable to run following tests, regardless of the subset of tests selected.
+TEST_COUNT=2
+
 test:
-       go test -v -timeout 10s -count 2
+       go test -v -timeout $(TEST_TIMEOUT) -count $(TEST_COUNT)
```

alban:
@pcmoore I updated the PR following your review.

I didn't squash the patches in order to retain the authorship of commits I've taken from the previous PR. But I can squash them into one patch if you prefer, let me know if I should do that for this PR :)

drakenclimber:
> Wow, 27 patches ... that's something. Especially considering I think you could count the associated main library patches to do enable userspace notification on one hand :)

Yeah, I have mixed feelings on this.  I think it's awesome to continue to give credit to the original authors, but some of the patches have extraneous noise in them that could cause issues later.

For  example this patch, b5b36ad032558b511cd60f198b070673d2e2290c, adds the notification APIs, but it also changes how the libseccomp C library is loaded and adds some erroneous whitespace to an otherwise unmodified file.  If for some reason, the library loading change is wrong and needed to be reverted, we couldn't simply revert this commit since it's interwined with the notifications API.

> I'm not sure yet if it is the case with this PR, but for future submissions please take the approach of simply updating flawed patches in the PR if the PR is still open and not merged. I generally dislike merging commits which are known to be broken; it both makes bisection difficult and the git log overly noisy.

Agreed.  Breaking git bisect (without a really good reason) is a risky thing to do.

alban:
I squashed the patches and updated the commitmsg. It should play nice with `git bisect` now :)

seccomp.go
```diff
@@ -34,19 +41,28 @@ type VersionError struct {
 	minimum string
 }
 
+// Caches the libseccomp API level
+var apiLevel uint
```
pcmoore, line 45: I really prefer for the various libseccomp language bindings to be as thin as possible, is there a reason why we need to have `apiLevel` cached in the golang bindings?  Was simply calling `getAPI()` as needed proving to be a problem in your application?

pcmoore (changes requested):

seccomp.go
```diff
@@ -540,6 +632,27 @@ func NewFilter(defaultAction ScmpAction) (*ScmpFilter, error) {
 	return filter, nil
 }
 
+// SetTsync sets or clears the filter's thread-sync (TSYNC) attribute. When set, this attribute
+// tells the kernel to synchronize all threads of the calling process to the same seccomp filter.
+// When using filters with the seccomp notification action (ActNotify), the TSYNC attribute
+// must be cleared prior to loading the filter. Refer to the seccomp manual page (seccomp(2)) for
+// further details.
+func (f *ScmpFilter) SetTsync(val bool) error {
```
pcmoore, line 640: Perhaps I'm reading this too quickly, but if we always enable the `TSYNC` bit in `NewFilter()` and there is little reason to disable the `TSYNC` bit in golang, why does the `SetTsync()` function exist?

pcmoore:
Thanks @alban, I appreciate your patience and persistence.  I realize this is likely a bit frustrating as we try to meet in the middle on this PR.

To be honest, I wasn't expecting all of the patches to be squashed down into one patch, but I recognize I probably didn't provide enough in the way of suggestions, so I'll take the blame for that (some basic guidance for the future provided below).  Beyond that, and the couple things I mentioned in GH comments above, I think it looks reasonable - please don't (re)split the patch.

However, moving forward, please use the following rules when organizing patches in a PR:
1. Each patch must compile and pass the test suite on it's own.  In other words you should be able to git-bisect into any patch inside this PR and the respository at that point should be functional.
2. Each patch should be significant on it's own.  Patches that simply do setup for later patches without improving anything themselves should be folded into those later patches.
3. If a later patch in the patchset invalidates something in an earlier patch, that portion of the earlier patch should be removed, so long as the other points made above are satisfied.

Hopefully this helps somewhat, but if you have any questions let me know.  Thanks.

seccomp.go
alban: When seccomp notification was initially implemented in Linux, it was initially implemented without support for `TSYNC`. See this commit that makes `NEW_LISTENER` and `TSYNC` flags exclusive, meaning the kernel rejects filters with both flags:
https://github.com/torvalds/linux/commit/7a0df7fbc14505e2e2be19ed08654a09e1ed5bf6

Then Linux 5.7 introduced a new API to enable seccomp notification with tsync together in this commit:
https://github.com/torvalds/linux/commit/51891498f2da78ee64dfad88fa53c9e85fb50abf

But for unfortunate users of Linux < 5.7, they need to disable `TSYNC` and use `runtime.LockOSThread()` instead to ensure that, upon forking, the new process retains the filter:
```
if seccompAPILevel == 5 {
  filter.SetTsync(false)
```

For users of Linux >= 5.7 with a sufficiently recent libseccomp (API Level >= 6), I agree that `SetTsync()` has little reason to exist. But I would like libseccomp-golang to work with older libseccomp and older kernels. And I think that the function `SetTsync()` unfortunately needs to be exposed to the users of the libseccomp-golang library so that they know they have to use `runtime.LockOSThread()` (it could cause incompatibility issues if the `TSYNC` flag was removed automatically by the library without explicit request).

seccomp.go
alban: I didn't have a problem with simply calling `getAPI()` but I wonder if Cesar had one. His initial commitmsg was:

> In addition, this change caches the libseccomp API level in variable "apiLevel" in order to avoid calling into libseccomp everytime we need to check the API level.

But the C function `seccomp_api_get()` already has a cache, so I don't see a reason to have a second cache in Golang, unless there was a problem with an old version of libseccomp we want to support. I am not aware of such a problem, so we can try to remove this cache.

ctalledo: Hi Alban, IIRC, caching apiLevel in Golang was merely an optimization to avoid calling getAPI() all the time. I don't believe there was a functional reason behind it. 

Thanks for driving this PR towards completion!

pcmoore: Okay, let's dtop the `apiLevel` caching and just stick with calls to `getAPI()` as needed.

seccomp.go
pcmoore: Ah ha, gotcha.  If you look at the main libseccomp library you'll see we struggled with this too.

With golang, I believe that if we have to choose between the `TSYNC` and notify functionality, `TSYNC` should win.  Instead of adding a new API and exposing something we really don't want to expose (disabling `TSYNC` for golang programs), I'd much rather restrict the notification functionality in the golang bindings to only those systems with an API level greater than or equal to six (supports the simultaneous use of `TSYNC` and notify).

Do you have any thoughts on this @drakenclimber?

alban: It sounds fine to me for what I want to do with runc.

For the history, we initially wrote [runc PR 2682](https://github.com/opencontainers/runc/pull/2682) using `SetTsync()` to be compatible with Linux < 5.7 (see [InitSeccomp()](https://github.com/opencontainers/runc/blob/1751ec3b27890e1e33ea3f23f61bd250b6b493b6/libcontainer/seccomp/seccomp_linux.go#L67-L82)) but we removed that in the last version of the runc PR (see [filterFlags()](https://github.com/opencontainers/runc/blob/c25bb2bc8556aee67824ac35a3224d12d4e50de6/libcontainer/seccomp/patchbpf/enosys_linux.go#L580-L586)) for keeping things simple, since we use Linux >= 5.7 anyway.

However, [nestybox/sysbox-runc](https://github.com/nestybox/sysbox-runc/) uses `SetTsync()` (see [prepNotify()](https://github.com/nestybox/sysbox-runc/blob/9d050e41cba7b32f43830b22c9a2bca97e01fa6f/libcontainer/seccomp/seccomp_linux.go#L311)) and might prefer to keep compatibility with older Linux distributions that use Linux < 5.7, which is possible to do in a safe way given how runc uses `runtime.LockOSThread()`.

cc @ctalledo

ctalledo: Hi @alban, @pcmoore,

Given that Linux supports seccomp-notify since earlier than 5.7, my preference would be to have seccomp-golang expose `SetTsync()`. But I understand that it's something that should ideally not be exposed as part of the seccomp-golang API, and that starting with Linux 5.7 there is less need to expose it (though IIRC calling `runtime.LockOSThread()` would still be needed). 

As far as the usage of seccomp-golang in [nestybox/sysbox-runc](https://github.com/nestybox/sysbox-runc), we definitely need sysbox-runc to support kernels < 5.7, as most users are on such kernels. Thus, if the consensus is that `SetTsync()` be removed from the seccomp-golang API, we would unfortunately need to continue using a custom version of seccomp-golang rather than the upstream version, until the need to support kernels < 5.7 goes away. 

Thanks again for all good work in driving this PR to completion, much appreciated!

ctalledo:
> I squashed the patches and updated the commitmsg. It should play nice with `git bisect` now :)

Apologies for the belated reply: thanks for squashing the commits as needed so `git bisect` works on them; when @rodnymolina and I initially made the commits, we had paid much attention to this, but I agree it's the proper way to do it.

seccomp.go
drakenclimber: > Ah ha, gotcha. If you look at the main libseccomp library you'll see we struggled with this too.
> 
> With golang, I believe that if we have to choose between the `TSYNC` and notify functionality, `TSYNC` should win. Instead of adding a new API and exposing something we really don't want to expose (disabling `TSYNC` for golang programs), I'd much rather restrict the notification functionality in the golang bindings to only those systems with an API level greater than or equal to six (supports the simultaneous use of `TSYNC` and notify).
> 
> Do you have any thoughts on this @drakenclimber?

Yes, this was real challenge in libseccomp.  Based on the comments above, I think we have to expose `SetTsync()` for legacy libseccomp-go users and users on older kernels.  But we should probably add a strong comment in it that warns of its dangers.

rata:
@pcmoore Hi, I'm a coworker of @alban! Sorry for the delay, we have some KubeCon deadlines now, but will address your comments in the coming weeks. Thanks for the review! :)

seccomp.go
rata: Added a big warning for this, then!

seccomp.go
rata: Dropped! :)

rata:
@pcmoore Hi! PTAL.

I've updated the branch to address the review comments. Here is a short summary of the changes:

* Fix SCMP_FLTATR_CTL_SSB with old libseccomp: this commit fixes a bug introduced recently in `main` to run with old libseccomp.
* tests: Add missing APILevelIsSupported() guard: Idem, fixes the tests with old libseccomp. This was broken recently in `main`
* Add seccomp userspace notification API to libseccomp-golang: This is the commit adding notify support, please note the commit msg was updated to reflect the state when the other commits are squashed to this.
* The rest: will be squashed to the "Add seccomp userspace notification API to libseccomp-golang" commit, but kept them separately to simplify review. These commit address the review comments and one other bug.

Please let me know if this looks fine now and I can squash after you review.

Thanks! :)

seccomp.go
pcmoore: Sorry all, I still think adding `SetTsync()` to the public API is a mistake.  My concern is that under Go it is far too easy to misuse this API and end up messing up the seccomp filtering for the process.  Further, the `SetTsync()` API is really only needed to workaround some limitations/bugs in a limited number of kernel releases; it's not necessary using a current or future kernels.

As mentioned above, I believe the best move here is to not expose `SetTsync()` and limit the notification functionality to only those systems with an API level greater than or equal to six.  Exposing a problematic API only to enable new functionality doesn't seem like a smart thing to do as far as I'm concerned.

rata: @pcmoore no problem, I've updated to remove `SetTsync()` now. Same notes as mentioned [here][link] apply. PTAL :)

[link]: https://github.com/seccomp/libseccomp-golang/pull/59#issuecomment-817791278

rata:
@pcmoore @drakenclimber friendly ping? :)

drakenclimber:
Sorry, last week was a busy one.  I will be doing libseccomp and libseccomp-golang stuff this week.

drakenclimber (approved):
Looks like the various outstanding issues from the previous review were resolved.  Thanks, @rata 

pcmoore:
Okay, let's merge this and we can fix up anything that breaks later.  Merged at 3879420cc921efb4f1a2d75bea84e6158ef21a78, thanks everyone!

ctalledo:
Yes, thanks to all that made the merge possible!

AkihiroSuda:
Can we have v0.9.2 with this soon?

rata:
@AkihiroSuda maybe you can create an issue for this? I see the 0.9.2 milestone exists and has more things to do, but will be great if a new release is out and other stuff are moved to 0.9.3 :)

pcmoore:
Two quick comments:

1. Please don't comment on closed issues unless your comment directly relates to the issue.  The comments above do not in my opinion relate to the original issue.
2. The obvious answer to helping get v0.9.2 released quicker is to help resolve the outstanding issues in the milestone; deferring issues to a later release is strongly discouraged and asking for that it not going to be received well by @drakenclimber or myself.

Issue #60, opened 2020-11-06
BUG: look into replacing Travis CI with GitHub actions


pcmoore:
There are plenty of articles on this, but see The Register article below for some background on this issue:

* https://www.theregister.com/2020/11/02/travis_ci_pricng

With Travis CI becoming potentially unreliable for libseccomp, we should investigate moving out CI testing to GitHub actions:

* https://docs.github.com/en/free-pro-team@latest/actions

drakenclimber:
Good find.  I have been running into issues with libcgroup running on TravisCI as well.  I'm very interested in this issue.

pcmoore:
Yes, I think most folks are starting to see problems with Travis CI.  I *think* we should be able to move to GH actions without too much problem but I haven't looked just yet, I may start playing with it this weekend depending on how things go around here.

With the main libseccomp v2.5.x releases stabilizing, we also need to start working through some of the issue/PR backlog here in the Go bindings.

kolyshkin:
PTAL #65.

drakenclimber:
Merged via commit https://github.com/seccomp/libseccomp-golang/commit/e5b1e63fb25d704a39b9079d758824587f31a5f2 and others.  CI is passing [here](https://github.com/seccomp/libseccomp-golang/runs/3570811836).

PR #65, opened 2021-08-03
RFE: add linters and enable GitHub CI

kolyshkin:
This PR
* fixes a few found linter warnings (all very minor)
* adds golangci-lint to makefile, replacing both golint (which is obsoleted) and govet (which is called by golangci-lint)
* enables some basic github CI checks for master branch and PRs:
  * golangci-lint
  * build and test checks for both supported golang versions (currently 1.16.x and 1.17.x) and the last unsupported one (1.15.x)

Fixes: #60

* v2 (Aug 19): added Go 1.17.x


kolyshkin:
Updated to add Go 1.17

kolyshkin:
@pcmoore @drakenclimber PTAL 🙏🏻 

This currently uses whatever libseccomp is available in Ubuntu-20.04. I can certainly add compile/test with libseccomp 2.5.2.

rata (approved):
As an outside contributor, this LGTM, thanks! I want to highlight that this should also close #67

pcmoore:
This looks okay to me, but we should hear from @drakenclimber, especially with respect to the GH Actions implementation.  Also, while not strictly necessary for this PR as it can be done later, we should probably go ahead and remove the Travis CI config while we are at it (if done in this PR, please do it as a separate patch).
```
```

pcmoore (approved):

seccomp.go
```diff
@@ -45,7 +45,7 @@ func init() {
 	// This forces the cgo libseccomp to initialize its internal API support state,
 	// which is necessary on older versions of libseccomp in order to work
 	// correctly.
-	GetAPI()
+	_, _ = GetAPI()
```
drakenclimber, line 48: This patch is no longer needed due to commit https://github.com/seccomp/libseccomp-golang/commit/ef015460ecc91e963a187dd865453cacb6f31899

drakenclimber (approved):
Looks good to me.  I'll merge the patchset in except for patch https://github.com/seccomp/libseccomp-golang/pull/65/commits/c2b914ede632014f5c7fde74ec85df40071971c3 which is already effectively committed.

We need to update the CI badge on the front page.  I added Issue https://github.com/seccomp/libseccomp-golang/issues/68 to address this.

`Reviewed-by: Tom Hromatka <tom.hromatka@oracle.com>`

pcmoore:
Thanks @drakenclimber, @kolyshkin, and @rata!

seccomp.go
kolyshkin: Rebased to remove this.

kolyshkin:
> Also, while not strictly necessary for this PR as it can be done later, we should probably go ahead and remove the Travis CI config while we are at it (if done in this PR, please do it as a separate patch).

done

kolyshkin:
Pushed a new version, rebasing on current tip. The only changes are:
* removed commit c2b914e (obsoleted by ef01546)
* added a commit that removes `.travis.yml`

kolyshkin:
> I can certainly add compile/test with libseccomp 2.5.2.

Do you think it makes sense to build/test against the latest libseccomp release, or whatever Ubuntu is providing (2.5.1 at the moment) is fine?

pcmoore:
> Do you think it makes sense to build/test against the latest libseccomp release, or whatever Ubuntu is providing (2.5.1 at the moment) is fine?

In general I would rather see upstream test against upstream as opposed to whatever happens to be the default in the CI.

drakenclimber:
Merged.  CI passed - see [here](https://github.com/seccomp/libseccomp-golang/runs/3570811836).

Let me know if you see any issues.

kolyshkin:
> In general I would rather see upstream test against upstream as opposed to whatever happens to be the default in the CI.
 
OTOH testing against whatever Ubuntu ships with make sense, too. Let me try to implement a matrix.

pcmoore:
> > In general I would rather see upstream test against upstream as opposed to whatever happens to be the default in the CI.
> 
> OTOH testing against whatever Ubuntu ships with make sense, too. Let me try to implement a matrix.

Let's stick to just upstream for now, I don't want to go off into the weeds just yet with distro testing.

PR #66, opened 2021-08-03
all: unify libseccomp version / API level checks

kolyshkin:
This unifies all the code that is used to check for minimally required
libseccomp version and API level and generate an error.

For errors, VersionError is reused, which is now amended with API level.

For checks, we now have checkVersion and checkAPI, which generate errors
that can be returned as is.

This removes some duplicated code and unifies error messages. As a
side benefit, a user can now check if an error was caused by an old
libseccomp and/or low API level, by using errors.As or checking if the
type is VersionError.

### DIFF
--- a/seccomp.go
+++ b/seccomp.go
@@ -34,32 +34,31 @@ import "C"
 
 // Exported types
 
-// VersionError denotes that the system libseccomp version is incompatible
-// with this package.
+// VersionError represents an error when either the system libseccomp version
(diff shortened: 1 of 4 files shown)


drakenclimber (approved):
Looks good to me.  Thanks @kolyshkin.

pcmoore:
Merged via ef015460ecc91e963a187dd865453cacb6f31899, thanks!

PR #67, opened 2021-08-31
Remove undeeded CI deps and update to libseccomp 2.5.1

rata:
Hi!

I found that some unneeded dependencies were inadvertently added by commit  78c92cbc6966350c27f365890f4cd47ebb934594, so I removed them. And while I was there, I realized that libseccomp 2.5.0 was being used and update it to use 2.5.1

I couldn't run the CI on my fork, travis is not liking it anymore for some reason (it used to work before), so I _think_ the CI will pass here but not 100% sure.

Let me know what you think

Best,
Rodrigo

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,8 +38,6 @@ addons:
   apt:
     packages:
       - build-essential
-      - astyle
-      - golint
       - gperf
 
 install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,10 +21,10 @@ language: go
 
 jobs:
   include:
-    - name: "last libseccomp 2.5.0"
+    - name: "last libseccomp 2.5"
       env:
-      - SECCOMP_VER=2.5.0
-      - SECCOMP_SHA256SUM=1ffa7038d2720ad191919816db3479295a4bcca1ec14e02f672539f4983014f3
+      - SECCOMP_VER=2.5.1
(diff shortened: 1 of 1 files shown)


rata:
Oh, it seems travis is not working here either :-/

I think travis was adding barriers to free projects, travis-ci.org seems to say that since June 2021 we should use travis-ci.com. Not sure if that is related, though.

kolyshkin:
> update to libseccomp 2.5.1

@rata 2.5.2 has been released recently

kolyshkin:
> Oh, it seems travis is not working here either :-/

This is fixed by #65; PTAL

rata:
@kolyshkin 
> This is fixed by #65; PTAL

Cool, thanks! That also addresses this, as the these deps I'm deleting are not there. Can you mark there that closes 67 too? :)

pcmoore:
Hi all, thanks for these fixes but given the problems with Travis CI and Open Source projects I think the right thing to do is to migrate to GitHub Actions (see PR #65).  For reference, we've moved the main libseccomp library to GH Actions and it seems to be working reasonably well.

Any objections if I close this PR without merging in favor of PR #65 and moving to GitHub Actions?

pcmoore:
I just saw @rata's comment in https://github.com/seccomp/libseccomp-golang/pull/65#pullrequestreview-751245279 so I'm going to close this out.  Thanks for your understanding and your review of PR #65!

Issue #68, opened 2021-09-10
BUG: Replace TravisCI Badge with Github Actions Badge


drakenclimber:

kolyshkin:
This should show the latest status for main branch:
[![validate](https://github.com/seccomp/libseccomp-golang/actions/workflows/validate.yml/badge.svg)](https://github.com/seccomp/libseccomp-golang/actions/workflows/validate.yml) [![test](https://github.com/seccomp/libseccomp-golang/actions/workflows/test.yml/badge.svg)](https://github.com/seccomp/libseccomp-golang/actions/workflows/test.yml)

pcmoore:
Closing this out as it is resolved.

Issue #40, opened 2020-03-13
RFE: evaluate README.md


pcmoore:
Compare the golang binding's README.md with the main libseccomp README.md to see if any additional edits should be made to the golang's README.md.  Whenever possible we should try to remove redundancy between the two, using the main libseccomp README.md as the main reference.

kolyshkin:
This is partially addressed by #69; frankly, I don't know what else to remove from there. An introductory paragraph about libseccomp seems very relevant and provide needed keywords for search engines. Other than that, it's all links to the docs.

pcmoore:
I think @kolyshkin's updates are enough to close this out, if anything remains we can address it via a new issue.  Thanks for your help @kolyshkin!

PR #69, opened 2021-09-10
RFE: README.md facelift

kolyshkin:
1. Remove travis, add GHA CI badge
2. Add pkg.go.dev badge
3. Remove old go/gcc version requirement
4. Dedup the testing notes, referring from README to CONTRIBUTING.

Please see individual commits for more details.

Fixes: #68
Addresses: #40

The badges part (result of the first two commits) should look like this:

[Go Reference](https://pkg.go.dev/github.com/seccomp/libseccomp-golang) [validate](https://github.com/seccomp/libseccomp-golang/actions/workflows/validate.yml) [test](https://github.com/seccomp/libseccomp-golang/actions/workflows/test.yml)

History:
* v2: also fixed doc link in the text
* v3: add the "dedup testing notes" commit

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,31 +1,23 @@
-How to Submit Patches to the libseccomp Project
+How to Submit Patches to the libseccomp-golang Project
 ===============================================================================
 https://github.com/seccomp/libseccomp-golang
 
 This document is intended to act as a guide to help you contribute to the
-libseccomp project.  It is not perfect, and there will always be exceptions
+libseccomp-golang project.  It is not perfect, and there will always be exceptions
 to the rules described here, but by following the instructions below you
 should have a much easier time getting your work merged with the upstream
 project.
 
 ## Test Your Code Using Existing Tests
 
-There are two possible tests you can run to verify your code.  The first
-test is used to check the formatting and coding style of your changes, you
-can run the test with the following command:
-
-	# make check-syntax
-
-... if there are any problems with your changes a diff/patch will be shown
-which indicates the problems and how to fix them.
-
-The second possible test is used to ensure the sanity of your code changes
-and to test these changes against the included tests.  You can run the test
-with the following command:
+A number of tests and lint related recipes are provided in the Makefile, if
+you want to run the standard regression tests, you can execute the following:
 
 	# make check
 
-... if there are any faults or errors they will be displayed.
+In order to use it, the 'golangci-lint' tool is needed, which can be found at:
+
+* https://github.com/golangci/golangci-lint
 
 ## Add New Tests for New Functionality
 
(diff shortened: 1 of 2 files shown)


kolyshkin:
This also shows how GHA CI look for a PR

CONTRIBUTING.md
```diff
@@ -1,31 +1,22 @@
...
-	# make check-syntax
-
-... if there are any problems with your changes a diff/patch will be shown
-which indicates the problems and how to fix them.
-
-The second possible test is used to ensure the sanity of your code changes
-and to test these changes against the included tests.  You can run the test
-with the following command:
+A number of tests and lint related recipes are provided in the Makefile, if
+you want to run the standard regression tests, you can excute the following:
```
drakenclimber, line 14: nitpick
`s/excute/execute`

drakenclimber (approved):
I found a small typo, but it looks good to me.  Thanks, @kolyshkin 

CONTRIBUTING.md
kolyshkin: Thanks! While it actually comes from commit 206feef4, I obviously overlooked it.

Updated to fix this.

CONTRIBUTING.md
```diff
@@ -1,31 +1,22 @@
...
-and to test these changes against the included tests.  You can run the test
-with the following command:
+A number of tests and lint related recipes are provided in the Makefile, if
+you want to run the standard regression tests, you can execute the following:
 
 	# make check
 
-... if there are any faults or errors they will be displayed.
+In order to use it, the 'golangci-lint' tool is needed, which can be found at:
+* https://github.com/golangci/golangci-lint
```
pcmoore, line 19: I will readily admit that this is *beyond* nitpicky, but are you okay with adding an empty line before the URL so it is more consistent with the other Markdown files in the project?  It doesn't make much of a difference when the Markdown is rendered, but I believe it increases readability when viewing the file in source form.

As this is such a minor tweak, if you are okay with the change we can add the blank line while merging the patches, just let us know if you would prefer that over an updated PR.

pcmoore:
Thanks @kolyshkin, this looks great.  I just had one small comment (see above), let me know what you would prefer to do.

CONTRIBUTING.md
kolyshkin: Totally justified and makes sense; commit updated.

pcmoore:
Thanks @kolyshkin, I just merged all the patches at HEAD 9da99da69b1b425f6e8ebb86fa61a656b84c6dcd.

As a quick FYI, when possible it's nice to try and keep the Markdown line length under 80 chars to make it easier to read in a terminal.  I should have noticed that sooner, but that doesn't always jump out at you when reviewing patches via the GH web UI, my apologies.  Regardless, it was a minor tweak to only a few lines so I just made the fix while merging the patches.  It's not a big deal, just something to keep in mind as you continue your work.

Thanks again!

PR #70, opened 2021-09-13
[WIP] CI/GHA: test against libseccomp 2.5.2 and HEAD

kolyshkin:
Currently, Ubuntu 20.04 (used in CI) ships libseccomp 2.5.1.
It surely makes sense to test against it, as this is what end users will
most probably end up using (on Ubuntu, that is -- other distros may
differ, of course, but currently many still use 2.5.1).

Yet, the latest released libseccomp is 2.5.2, and it makes sense to test
against it as well, so add it to the matrix.

While at it, it's relatively easy to enable testing against git tip as
well, so do it.


drakenclimber:
Thanks @kolyshkin.  A matrix seemed like a good way to meet everyone's testing needs; I'm glad you added it.

`Reviewed-by: Tom Hromatka <tom.hromatka@oracle.com>`

drakenclimber (approved):

.github/workflows/test.yml
```diff
@@ -16,23 +16,54 @@ jobs:
       fail-fast: false
       matrix:
         go-version: [1.15.x, 1.16.x, 1.17.x]
+        libseccomp: ["", "v2.5.2", "HEAD"]
```
pcmoore, line 19: Thanks for doing this @kolyshkin, but as mentioned on another issue/PR somewhere, I'm not a big fan of testing against the default Ubuntu package as I don't want to see our CI failing because a distro made a mistake with the packaging (mistakes are our job! <g>).

I understand the desire to test against downstream distros, but I would much rather see us remain as distro-agnostic as possible and focus on the upstream bits when doing upstream testing.

drakenclimber: Seems like a reasonable argument to me.  I think the matrix still makes sense, but obviously it would become
`libseccomp: ["v2.5.2", "HEAD"]`

pcmoore: Yes, the matrix testing is really nice, I don't want to lose that, I just don't want to bother with testing and gating on Ubuntu's package.

kolyshkin: I think testing against whatever a distro is providing adds value, as personally I'd like to know if it fails. Besides, what we use here in CI is a stable distro, not some bleeding edge daily build or something, and we already depend on it a lot (glibc/gcc/gperf/binutils and so on).

It seems that issue you refer to can be solved by not treating all the CI jobs equal, but rather classify them into "required" or "optional". If you and this can surely be placed into the latter category (same is true for test against git tip, which IMO have much higher chance of breakage).

Anyway, it's easy to implement; done. Please let me know if you think it makes sense or not to explicitly add 2.5.1 (compiled from source) to the matrix (I tend to think it's not).

kolyshkin:
Looking more into libseccomp-golang, it seems that it will benefit from testing against older supported libseccomp releases, not just the latest. Currently the minimum supported version is 2.2.0, let's try to add it.

.github/workflows/test.yml
kolyshkin: Ended up having various minor supported libseccomp releases, as this is what we had in travis.

kolyshkin:
Nice -- added libseccomp 2.3.x and 2.4.x revealed a bug in TestNotif -- it checks the API level but do not check libseccomp version :)

Will fix later today, changing to DRAFT for now.

kolyshkin:
Looks like this is finally ready; took me a while to fix all the issues found 😅

kolyshkin:
Ideally, I think, it makes sense to cache the result of libseccomp compile (and avoid git clone/configure/make install if the cache is sufficient), but this can definitely be implemented separately on top of this.

kolyshkin:
> @drakenclimber drakenclimber added the enhancement label 5 days ago

With all the fixes added to support older libseccomp version, I'd characterize this as bugfix rather than enhancement now.

kolyshkin:
@drakenclimber @pcmoore PTAL.

If sending patches to the ML is preferable, please let me know (currently, CONTRIBUTING.md says otherwise).

drakenclimber:
> If sending patches to the ML is preferable, please let me know (currently, CONTRIBUTING.md says otherwise).

I personally prefer github pull requests.  But I wear many hats and unfortunately libseccomp-golang isn't one of them this week.  The Linux Plumbers Conference is currently going on all week.

pcmoore:
> > If sending patches to the ML is preferable, please let me know (currently, CONTRIBUTING.md says otherwise).
> 
> I personally prefer github pull requests. But I wear many hats and unfortunately libseccomp-golang isn't one of them this week. The Linux Plumbers Conference is currently going on all week.

+1 to what @drakenclimber already said.  This has been a particularly difficult week on my end due to some personal issues, LPC, and a handful of high-priority and high-visibility SELinux kernel issues flaring up.  I'm working my way through the libseccomp-golang PRs right now ...

.github/workflows/test.yml
```diff
@@ -16,7 +16,7 @@ jobs:
       fail-fast: false
       matrix:
         go-version: [1.15.x, 1.16.x, 1.17.x]
-        libseccomp: ["v2.5.2", "HEAD"]
+        libseccomp: ["v2.2.1", "v2.2.3", "v2.3.3", "v2.4.3", "v2.5.2", "HEAD"]
```
pcmoore, line 19: This is interesting as we no longer support v2.2.x, v2.3.x, or v2.4.x upstream anymore.  I guess we can let this slide for no so long as there are no major issues, but as soon as something significant crops up with these older releases we should probably just drop them.

pcmoore:
Thanks again for the patches @kolyshkin, but I have a few general comments, in no particular order:

* I dislike combining unrelated patches into a single PR simply because while working on item A you happened to stumble across item B.  I get it, it's *extremely* tempting to just do everything in one PR, but it is a bit of a mess to review.  I'd much rather you see things where you post one PR, get it merged, and then post the other; or post both PRs at the same time with a comment that PR A is dependent on PR b (or something similar).

* Your patch prefixes are not consistent, which I guess is somewhat understandable given the repo and what I'm used to seeing with golang.  When I say not consistent I'm talking about mixing function names, filenames, wildcards, etc.  I very much prefer using patch subject prefixes that are either a filename or a subsystem name, e.g. "tests: test all the things" or "seccomp_internal: bump the minimum supported version to X.Y..Z"; they are less transient than function names.  Please don't ever use wildcards in the subject prefix.

* Be careful when you talk about "supported versions" in the commit descriptions and code comments as there is ambiguity around who is providing support.  For example, there are distros which are still providing user support but are shipping libseccomp v2.3 (possibly older) and we/upstream don't currently support releases below v2.5.

My apologies if all of the above seem a bit trivial and nit-picky, but as you are starting to contribute more - thank you! - you're going to be subjected to a bit more "tough love" if you will :)  Please leave this PR as-is, I'll fix these things up when I hear back from @drakenclimber (below), but it is something to keep in mind for the future.

@drakenclimber, this PR has changed quite a bit since you last looked at it, does your `Reviewed-by` tag still stand?

.github/workflows/test.yml
kolyshkin: I'm merely following what's in there already (except for dropping 2.2.0 support).

If you want, I can rework this to drop 2.2.x support, too, as this is only used by very old (but still supported) Ubuntu 14.04. 
Dropping 2.3 and especially 2.4 is more problematic though, as this is what many long-term supported distros ship (Debian, RHEL, SLES).

I think it makes total sense to drop 2.2 and 2.3 (and maybe 2.4) _after_ the release is cut, but it's up to you to decide.

kolyshkin:
> I dislike combining unrelated patches into a single PR simply because while working on item A you happened to stumble across item B. I get it, it's extremely tempting to just do everything in one PR, but it is a bit of a mess to review. I'd much rather you see things where you post one PR, get it merged, and then post the other; or post both PRs at the same time with a comment that PR A is dependent on PR b (or something similar).

If you got an impression that I piled up everything I can change into this single PR, let me assure you this is not so. It all started with "let's add libseccomp versions that this package seems to support to CI", and the rest was added to make the CI green.

The only non-really-related stuff is "seccomp_internal.go: simplify version checks" and "test: execInSubprocess: simplify", and I will move it out of this commit.

> Be careful when you talk about "supported versions" in the commit descriptions and code comments 

Will reword.

> My apologies if all of the above seem a bit trivial and nit-picky

Definitely no need to apologize, any constructive feedback (and "tough love") is very welcome and truly appreciated.

> Please leave this PR as-is.

No problem, I'd rather fix it in place.

kolyshkin:
Rebased on top of the current main

kolyshkin:
@drakenclimber @pcmoore PTAL 🙏🏻

pcmoore:
Patience please. This week, and last week for that matter, have been busy with conferences as well as the usual stuff. On a personal note, the late summer, early fall weather has been really nice where I live too so I've been trying to enjoy that outside of work.

We see the activity on this, and appreciate it, but please be patient and allow us the time to review it.

drakenclimber:
Thanks, @kolyshkin.  The most recent changes look good to me.

drakenclimber (approved):

pcmoore:
Merged at HEAD 6a935f5222db4e0543c15142f4ae9d3830d1d671, thanks @kolyshkin!

PR #74, opened 2021-09-17
AddRule: fix to handle EACCES

kolyshkin:
In case a rule with the action that equals to the default one was added,
libseccomp used to return EPERM, and libseccomp-golang converted it into
a more user-friendly "requested action matches default action of
filter" error.

From various bug reports I have noticed this is no longer a case.

The cause is https://github.com/seccomp/libseccomp/commit/83989be02 (appeared in v2.5.0), which
changes EPERM to EACCES.

Since we still support libseccomp < 2.5.0, check for either EPERM or
EACCES, and add a TODO item to remove the former.

Add a test case, which fails like this before the fix:

> seccomp_test.go:580: expected error to contain "matches default action", got "permission denied"


seccomp_test.go
```diff
@@ -559,6 +559,28 @@ func TestMergeFilters(t *testing.T) {
...
+	if err != nil {
+		t.Errorf("Error creating filter: %s", err)
+	}
+	defer filter.Release()
+
+	err = filter.AddRule(ScmpSyscall(0x1), ActAllow)
+	if err == nil {
+		t.Error("expected error, got nil")
+	} else {
+		const exp = "matches default action"
```
kolyshkin, line 577: Alternatively, we can introduce
```go
var errRuleActionMatchesDefault = errors.New("requested action matches default action of filter")
```
in the code, and compare against it here in the test. This way, if the wording is ever changed, the test would not need fixing (but will still check the error returned is as expected).

kolyshkin: Implemented.

pcmoore:
Hmmm, I think @drakenclimber and I may need to sort an aspect of this out in the main libseccomp library first as the `seccomp_rule_add(3)` function returning `-EACCESS` or `-EPERM` is not documented as part of our stable API promise.

Either we need to update the docs (the manpage) or we need to shift to using one of the approved error codes, `-EEXIST` seems the most appropriate since the rule is conflicting with the default action/rule.

pcmoore:
As I look at this I continue to be a bit torn, `-EACCESS` does make sense in the main libseccomp API for `seccomp_rule_add(3)` based on what we use it for in other funcs in the libseccomp API.  Curious to hear what you think @drakenclimber.

drakenclimber:
Interesting.  For good or for bad, libseccomp proper is consistent with error codes (`-EACCES`) around the default action.  I haven't thought deeply about it, but `-EPERM` feels more intuitive to me.  I could see someone misinterpreting `-EACCES` as insufficient privileges.

Paul is correct in that we need to fix something (code and/or manpages) in libseccomp.  Late on a Friday of a long week, I would vote for returning `-EPERM` and updating the manpage accordingly.  But I don't feel strongly about it... and it's getting late on Friday :)

kolyshkin:
> the seccomp_rule_add(3) function returning -EACCESS or -EPERM is not documented as part of our stable API promise

... yet the commit https://github.com/seccomp/libseccomp/commit/83989be02 I refer to in description says "[t]his is part of our error code cleanup and API promise".

> I haven't thought deeply about it, but `-EPERM` feels more intuitive to me. I could see someone misinterpreting `-EACCES` as insufficient privileges.

My take on that is, since either EACCES or EPERM can be misinterpreted, let's stick with the one we already have (since 2.5.0), i.e. EACCES. Changing it back to EPERM, it seems, will do more harm than good.

> Paul is correct in that we need to fix something (code and/or manpages) in libseccomp.  

I agree. Filed https://github.com/seccomp/libseccomp/pull/339 so we can continue discussion in there.

Moving this PR to draft for now.

kolyshkin:
No longer a draft.

PR #77, opened 2021-09-25
RFE: simplify version checks

kolyshkin:
As the code errors out earlier if seccomp version is less than 2.2.0,
there is no need to check for SCMP_VER_MAJOR < 2.

Drop it.


pcmoore (approved):
Yeah, the golang bindings aren't even remotely compatible with libseccomp v1.x :)
```
```

pcmoore:
I talked with @drakenclimber about this offline and he's okay with merging it so I'm going to close this out.  Merged via 398b733fa47fc9618576257e319f184bfc394996.

PR #85, opened 2022-01-28
BUG: define ActKillThread equal to ActKill

tonistiigi:
These constants are equal in libseccomp https://github.com/seccomp/libseccomp/commit/b2f15f3d02f302b12b9d1a37d83521e6f9e08841#diff-26b243770fb00079467cda2a0d84f4d3d4158e565eac4570f12b08cbfe172be2R255 (with ActKillThread being a more precise name for the previous behavior)  but Go definitions were defined separately in https://github.com/seccomp/libseccomp-golang/commit/24f29379854785de68bbd44630cc093b22b4dc29.

This resulted in [dead code](https://github.com/seccomp/libseccomp-golang/commit/24f29379854785de68bbd44630cc093b22b4dc29#diff-e4bc90dec05206f6ee3b16a8b0bb5a6a6b89face1b97420f5b4d437ee12f7cf5R539-R540) that never executed due to identical case statements in switch. Go can [usually](https://go.dev/play/p/QHFMuzVfW0o) detect these error cases and refuses to build but for some reason this detection doesn’t work with cgo+gcc.

Clang detects the equal constants correctly and therefore `libseccomp-golang` builds with clang broke after `ActKillThread` was added making it impossible for us to build the latest `runc`. https://gist.github.com/tonistiigi/ab6ef71781b7e3c5cabcb71a286a9243 is a simple example showing this condition where builds with clang fail.

In order to fix the clang build only removal of the switch case is needed. But I assumed that [the setter/getter
logic](https://github.com/seccomp/libseccomp-golang/blob/2a71848205430267e34457b3b771a37e7d584956/seccomp_test.go#L405-L416) is supposed to work for `ActKillThread` as well and the only way to ensure that is to set them equal like they
are in C.

@kolyshkin @pjbgf

Signed-off-by: Tonis Tiigi <tonistiigi@gmail.com>

### DIFF
--- a/seccomp.go
+++ b/seccomp.go
@@ -202,14 +202,14 @@ const (
 	// This action is only usable when libseccomp API level 3 or higher is
 	// supported.
 	ActLog
-	// ActKillThread kills the thread that violated the rule. It is the same as ActKill.
-	// All other threads from the same thread group will continue to execute.
-	ActKillThread
 	// ActKillProcess kills the process that violated the rule.
 	// All threads in the thread group are also terminated.
 	// This action is only usable when libseccomp API level 3 or higher is
 	// supported.
 	ActKillProcess
+	// ActKillThread kills the thread that violated the rule. It is the same as ActKill.
+	// All other threads from the same thread group will continue to execute.
+	ActKillThread = ActKill
 )
 
 const (
@@ -385,7 +385,7 @@ func (a ScmpCompareOp) String() string {
 // String returns a string representation of a seccomp match action
 func (a ScmpAction) String() string {
 	switch a & 0xFFFF {
-	case ActKill, ActKillThread:
+	case ActKillThread:
 		return "Action: Kill thread"
 	case ActKillProcess:
 		return "Action: Kill process"
--- a/seccomp_internal.go
+++ b/seccomp_internal.go
@@ -629,8 +629,6 @@ func (a ScmpCompareOp) toNative() C.int {
 func actionFromNative(a C.uint32_t) (ScmpAction, error) {
 	aTmp := a & 0xFFFF
 	switch a & 0xFFFF0000 {
-	case C.C_ACT_KILL:
-		return ActKill, nil
 	case C.C_ACT_KILL_PROCESS:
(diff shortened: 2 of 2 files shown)


thaJeztah:
@kolyshkin ptal 🤗

pjbgf (approved):
Nice catch @tonistiigi. LGTM

kolyshkin (approved):
LGTM.

I think it also makes sense to mark `ActKill` as deprecated, asking to use `ActKillThread` instead (and changing the code accordingly, i.e. `s/ActKill/ActKillThread/`). Surely this can be done separately.

drakenclimber (approved):

seccomp.go
```diff
@@ -202,14 +202,14 @@ const (
...
-	// All other threads from the same thread group will continue to execute.
-	ActKillThread
 	// ActKillProcess kills the process that violated the rule.
 	// All threads in the thread group are also terminated.
 	// This action is only usable when libseccomp API level 3 or higher is
 	// supported.
 	ActKillProcess
+	// ActKillThread kills the thread that violated the rule. It is the same as ActKill.
+	// All other threads from the same thread group will continue to execute.
+	ActKillThread = ActKill
```
pcmoore, line 212: This may be very nitpicky, but if we are trying to be similar to the core libseccomp library it may be better to do:
```
ActKill = ActKillThread
```
... or would that still cause problems with clang?

pcmoore: I tend to think it is in scope, if we are changing how we define `ActKillThread` let's go ahead and do things properly so we are better in sync with the core library.  A quick grep of the libseccomp-golang sources shows approximately two dozen places in the code that might be affected; that's not bad as far as I'm concerned.

I'm fine if you want to do this work in a separate patch, but please keep it in this PR.

thaJeztah:
To address https://github.com/seccomp/libseccomp-golang/pull/85#discussion_r813429461, I opened https://github.com/seccomp/libseccomp-golang/pull/90 which swaps `ActKillThread` and `ActKill`, and marks `ActKill` as deprecated.

Issue #36, opened 2020-03-12
ADMIN: change maintainers of the libseccomp-golang project


pcmoore:
With @mheon stepping down (see [this comment](https://github.com/seccomp/libseccomp-golang/pull/35#issuecomment-582433868) in PR #35) let's shift maintenance of the the libseccomp golang bindings to @drakenclimber and myself.  The checklist below is a work in progress (expect it to be edited while this issue remains open) but should be used to guide this work.

Work Items:

- [x] Ensure @drakenclimber and @pcmoore have the appropriate repo permissions on GitHub
- [x] Remove @mheon's write permissions
- [x] Convert README to Markdown
- [x] Convert SUBMITTING_PATCHES to Markdown and rename to CONTRIBUTING.md
- [x] Add a SECURITY.md guidance doc
- [x] Add a doc/admin/MAINTAINER_PROCESS.md guidance doc
- [x] Add a doc/admin/RELEASE_PROCESS.md guidance doc.  Add a step to ensure the [test matrix](https://github.com/seccomp/libseccomp-golang/pull/70/commits/59c82f9ca6ff40c91a11e689bb1c68c7c3c9e8c8#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R19) is updated.
- [x] Triage all outstanding issues and PRs (ensure proper labeling, status, etc.)
- [x] Add @kolyshkin as a libseccomp-golang maintainer (see #87)
- [ ] Assess current milestones and release plans

pcmoore:
@drakenclimber I know we talked briefly about this over email a while back, but I'm roping you into this now like it or not :)  It's probably a good thing as it will mean tighter integration between the bindings and the main library, but it does mean we will need to get a bit more versed in golang (also not an entirely bad thing).

Comments are welcome, but objections to your new role will be ignored ;)

pcmoore:
Regarding docs, we should update the existing README (also convert it to Markdown), SUBMITTING_PATCHES (Markdown conversion, rename to CONTRIBUTING.md so the GH magic works better) and add in a SECURITY.md and doc/admin/MAINTAINER_PROCESS.md and doc/admin/RELEASE_PROCESS.md.

The main libseccomp docs should be a good starting point, but those docs will need to be re-examined in the context of golang conventions and this project.

drakenclimber:
Sounds good.  I can help with the update.

Your checklist looks like a really good starting point.

[...]

PR #86, opened 2022-02-02
RFE: add admin docs as well as a local copy of SECURITY.md

pcmoore:
This are based on the core libseccomp docs.

Signed-off-by: Paul Moore <paul@paul-moore.com>

### DIFF
--- a/doc/admin/MAINTAINER_PROCESS.md
+++ b/doc/admin/MAINTAINER_PROCESS.md
@@ -1,0 +1,98 @@
+The libseccomp-golang Maintainer Process
+===============================================================================
+https://github.com/seccomp/libseccomp-golang
+
+This document attempts to describe the processes that should be followed by the
+various libseccomp-golang maintainers.  It is not intended as a hard
+requirement, but rather as a guiding document intended to make it easier for
+multiple co-maintainers to manage the libseccomp-golang project.
+
+We recognize this document, like all other parts of the libseccomp-golang
+project, is not perfect.  If changes need to be made, they should be made
+following the guidelines described here.
+
+### Reviewing and Merging Patches
+
+In a perfect world each patch would be independently reviewed and ACK'd by each
+maintainer, but we recognize that is not likely to be practical for each patch.
+Under normal circumstances, each patch should be ACK'd by a simple majority of
+maintainers (in the case of an even number of maintainers, N/2+1) before being
+merged into the repository.  Maintainers should ACK patches using a format
+similar to the Linux Kernel, for example:
+
+```
+Acked-by: John Smith <john.smith@email.org>
+```
+
+The maintainer which merged the patch into the repository should add their
+sign-off after ensuring that it is correct to do so (see the documentation on
+submitting patches); if it is not correct for the maintainer to add their
+sign-off, it is likely the patch should not be merged.  The maintainer should
+add their sign-off using the standard format at the end of the patch's
+metadata, for example:
+
+```
+Signed-off-by: Jane Smith <jane.smith@email.org>
+```
+
+The maintainers are encouraged to communicate with each other for many reasons,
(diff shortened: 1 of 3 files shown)


pcmoore:
This ticks off a few more boxes in issue #36.

doc/admin/MAINTAINER_PROCESS.md
```diff
@@ -0,0 +1,98 @@
...
+Feature requests MUST have a "RFE:" prefix added to the issue name and use the
+"enhancement" label.  Bug reports MUST a "BUG:" prefix added to the issue name
+and use the "bug" label.
+
+Issues SHOULD be prioritized using the "priority/high", "priority/medium", and
+"priority/low" labels.  The meaning should hopefully be obvious.
+
+Issues CAN be additionally labeled with the "pending/info", "pending/review",
+and "pending/revision" labels to indicate that additional information is
+needed, the issue/patch is pending review, and/or the patch requires changes.
```
kolyshkin, line 68: I'm not sure that `pending/review` and `pending/revision` labels are needed.

If this is about issues, `pending/review` might make sense (although I was not able to find any issue, either opened of closed, with that label in the seccomp/libseccomp repo). As for `pending/revision`, for issues its meaning is similar to `pending/info`, which essentially asks the reporter to amend the issue in some way. This label is also unused for issues in seccomp/libseccomp.

If this is about patches (pull requests in GH lingo), then `pending/review` is a natural initial state of an opened PR, and IMO does not require a specific label. A reviewer may choose `Request changes` when submitting a review, and GH shows this state, thus having `pending/revision` would be a duplicate of the existing review functionality.

Perhaps this text was written before GH added a review functionality. In any case, I guess it needs to be revised (in seccomp/libseccomp main repo, not here, since this is mostly a copy of text from that repo).

pcmoore: > I'm not sure that `pending/review` and `pending/revision` labels are needed.

This why the `pending/*` labels are marked as "CAN" not "MUST" or "SHOULD".  I personally find them helpful but I recognize that not everyone does and as a result I don't want to force them on anyone.  However, they are described here, albeit in very broad terms, for those who do want to make use of them.

> If this is about issues, `pending/review` might make sense (although I was not able to find any issue, either opened of closed, with that label in the seccomp/libseccomp repo). As for `pending/revision`, for issues its meaning is similar to `pending/info`, which essentially asks the reporter to amend the issue in some way. This label is also unused for issues in seccomp/libseccomp.

The line between issues and PRs in GH is a bit fuzzy (although they have been slowly growing a bit more distinct over time); while I generally tend to use these labels only for PRs, they could also be used for issues.  See my comment above about the `pending` labels and "CAN".

It is also worth noting that in many cases I tend to clear these `pending` labels when closing the issue/PR, so please don't read anything into the presence of these labels in a repo.  Also, see my "CAN" comments :)

> If this is about patches (pull requests in GH lingo), then `pending/review` is a natural initial state of an opened PR, and IMO does not require a specific label. A reviewer may choose `Request changes` when submitting a review, and GH shows this state, thus having `pending/revision` would be a duplicate of the existing review functionality.

The `pending` labels are also a nice indicator that someone has looked at the issue/PR and at least done a basic triage of the issue, even if the triage is simply `pending/review`.

drakenclimber (approved):
Looks good to me.  Thanks, @pcmoore 

doc/admin/MAINTAINER_PROCESS.md
drakenclimber: I don't usually use the `pending` labels, but I recognize they're helpful to others.  I'm fine with whatever direction you choose here.

pcmoore:
Merged via e214ef109e1074a6e8b06d4d21b422b10712a2ed, thanks for the reviews everyone!

Issue #89, opened 2022-02-20
BUG: unable to compile on Archlinux


quinnjr:
Attempts to compile this library on Archlinux for using `snap` result in the following error:

```sh
# github.com/snapcore/snapd/vendor/github.com/seccomp/libseccomp-golang
../go/src/github.com/snapcore/snapd/vendor/github.com/seccomp/libseccomp-golang/seccomp_internal.go:630:2: duplicate case _Ciconst_C_ACT_KILL_THREAD (value 0) in switch
	previous case at ../go/src/github.com/snapcore/snapd/vendor/github.com/seccomp/libseccomp-golang/seccomp_internal.go:626:2
```

The library does compile correctly in a clean chroot. The only thing I can really find different in my system versus a clean chroot is that I have CC/CXX set to Clang versus gcc.

kolyshkin:
I believe this is fixed by https://github.com/seccomp/libseccomp-golang/pull/85, which is currently being reviewed.

pcmoore:
With #85 merged, any chance @quinnjr that you can verify that the current main branch solved your problem?

quinnjr:
> With #85 merged, any chance @quinnjr that you can verify that the current main branch solved your problem?

No, still having the same issue with the switch statement.

```
*** Setting version to '2.55.4-1' from user.
# github.com/snapcore/snapd/vendor/github.com/seccomp/libseccomp-golang
../go/src/github.com/snapcore/snapd/vendor/github.com/seccomp/libseccomp-golang/seccomp_internal.go:630:7: duplicate case (_Ciconst_C_ACT_KILL_THREAD) (constant 0 of type C.uint) in expression switch
	../go/src/github.com/snapcore/snapd/vendor/github.com/seccomp/libseccomp-golang/seccomp_internal.go:626:7: previous case
```

pcmoore:
Hmm, I wonder if this is due to something related to snap?  On my current (as of May 8, 2022) Arch system with my libseccomp-golang repo sitting at f57e1d55ea18cbace22a9eefbd5e83417a2c1ace, I'm able to build and run the tests without problem.
```
% make test
go test -v -timeout 10s
=== RUN   TestExpectedSeccompVersion
...
ok      github.com/seccomp/libseccomp-golang    0.021s
```

@quinnjr are you able to fetch the current repo and successfully do a `make test` outside of a snap environment?

quinnjr:
[...]

PR #90, opened 2022-04-22
BUG: define ActKillThread equal to ActKill, and deprecated ActKill

thaJeztah:
- Carry of / closes https://github.com/seccomp/libseccomp-golang/pull/85 (second commit is new)
- relates to / addresses https://github.com/seccomp/libseccomp-golang/issues/89
- relates to / addresses https://github.com/moby/buildkit/pull/2584

> These constants are [equal in libseccomp][1], with ActKillThread being a more
> precise name for the previous behavior, but Go definitions were defined separately
> in 24f29379854785de68bbd44630cc093b22b4dc29.
> 
> This resulted in [dead code][2] that never executed due to identical case statements
> in switch. Go can [usually][3] detect these error cases and refuses to build but
> for some reason this detection doesn’t work with cgo+gcc.
> 
> Clang detects the equal constants correctly and therefore `libseccomp-golang` builds
> with clang broke after `ActKillThread` was added making it impossible for us to
> build the latest `runc`. https://gist.github.com/tonistiigi/ab6ef71781b7e3c5cabcb71a286a9243
> is a simple example showing this condition where builds with clang fail.
> 
> In order to fix the clang build only removal of the switch case is needed. But I
> assumed that [the setter/getter logic][4] is supposed to work for `ActKillThread`
> as well and the only way to ensure that is to set them equal like they are in C.

[1]: https://github.com/seccomp/libseccomp/commit/b2f15f3d02f302b12b9d1a37d83521e6f9e08841#diff-26b243770fb00079467cda2a0d84f4d3d4158e565eac4570f12b08cbfe172be2R255
[2]: https://github.com/seccomp/libseccomp-golang/commit/24f29379854785de68bbd44630cc093b22b4dc29#diff-e4bc90dec05206f6ee3b16a8b0bb5a6a6b89face1b97420f5b4d437ee12f7cf5R539-R540
[3]: https://go.dev/play/p/QHFMuzVfW0o
[4]: https://github.com/seccomp/libseccomp-golang/blob/2a71848205430267e34457b3b771a37e7d584956/seccomp_test.go#L405-L416

### DIFF
--- a/seccomp.go
+++ b/seccomp.go
@@ -202,14 +202,14 @@ const (
 	// This action is only usable when libseccomp API level 3 or higher is
 	// supported.
 	ActLog
-	// ActKillThread kills the thread that violated the rule. It is the same as ActKill.
-	// All other threads from the same thread group will continue to execute.
-	ActKillThread
 	// ActKillProcess kills the process that violated the rule.
 	// All threads in the thread group are also terminated.
 	// This action is only usable when libseccomp API level 3 or higher is
 	// supported.
 	ActKillProcess
+	// ActKillThread kills the thread that violated the rule. It is the same as ActKill.
+	// All other threads from the same thread group will continue to execute.
+	ActKillThread = ActKill
 )
 
 const (
@@ -385,7 +385,7 @@ func (a ScmpCompareOp) String() string {
 // String returns a string representation of a seccomp match action
 func (a ScmpAction) String() string {
 	switch a & 0xFFFF {
-	case ActKill, ActKillThread:
+	case ActKillThread:
 		return "Action: Kill thread"
 	case ActKillProcess:
 		return "Action: Kill process"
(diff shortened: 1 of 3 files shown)


thaJeztah:
@tonistiigi @pcmoore @kolyshkin PTAL (I see there's no CI running in this repo, so could use the eyes to double-check this is good)

thaJeztah:
oh, never mind; I was typing too fast; it's just waiting approval to run 😂

crazy-max (approved):
LGTM

thaJeztah:
oh, let me ping @drakenclimber as well (I see I forgot) as they reviewed the original PR; ptal 🤗

kolyshkin (approved):
LGTM, and also addresses https://github.com/seccomp/libseccomp-golang/pull/85#discussion_r812454715

drakenclimber (approved):
Looks good to me.  Thanks @thaJeztah

pcmoore:
Merged at HEAD f33da4d89646b6c01cb104d81094208b83a9b8e4, thank you!

thaJeztah:
Thank you!

Issue #87, opened 2022-02-10
ADMIN: add Kir Kolyshkin as a maintainer


pcmoore:
We (@pcmoore and @drakenclimber) have asked @kolyshkin if he would like to join the libseccomp-goland project as a co-maintainer and he graciously agreed (thanks Kir!).  This issue is intended to track all of the different things we need to do to ensure Kir is setup properly as a maintainer for the libseccomp-golang bindings.

*NOTE: the list items below are subject to change based on the discussion in this issue.*

- [X] Verify that @kolyshkin has 2FA enabled for his GH account
- [x] @kolyshkin should setup a PGP/GPG key if he hasn't already for tag/release signing (RELEASE_PROCESS.md is still a TODO)
- [x] Update issue #36 with @kolyshkin's account
- [x] Check and update README.md, CONTRIBUTING.md, SECURITY.md, and MAINTAINER_PROCESS.md (pending PR) if needed with @kolyshkin's info
- [x] Grant @kolyshkin write access to seccomp/libseccomp-golang

pcmoore:
@drakenclimber and @kolyshkin, please suggest any edits or new list items for the checklist above.

... and thanks again @kolyshkin for agreeing to help out!

pcmoore:
Removed the mailing list item from the list above as the mailing list has been deprecated.

kolyshkin:
>  @kolyshkin should setup a PGP/GPG key if he hasn't already for tag/release signing (RELEASE_PROCESS.md is still a TODO)

I have an RSA2048 key (which is probably not good enough these days but I'm lazy), see e.g.
https://keyserver.ubuntu.com/pks/lookup?op=vindex&search=0x17de5ecb75a1100e

> Check and update README.md, CONTRIBUTING.md, SECURITY.md, and MAINTAINER_PROCESS.md (pending PR) if needed with @kolyshkin's info

I took a look and the only place I see to which my name can/should be added is SECURITY.md. Here's a PR: https://github.com/seccomp/libseccomp-golang/pull/93

pcmoore:
> I have an RSA2048 key (which is probably not good enough these days but I'm lazy), see e.g. https://keyserver.ubuntu.com/pks/lookup?op=vindex&search=0x17de5ecb75a1100e

[...]

PR #91, opened 2022-05-02
RFE: bump the versions in the GH actions

pcmoore:
According to the Go release policy[1], only the two most recent releases are supported:

> Each major Go release is supported until there are two newer major releases. For example, Go 1.5 was supported until the Go 1.7 release, and Go 1.6 was supported until the Go 1.8 release. We fix critical problems, including critical security problems, in supported releases as needed by issuing minor revisions (for example, Go 1.6.1, Go 1.6.2, and so on).

... and with Go v1.18 being the most recent major release, trim the Go versions to the two supported 1.18.x and 1.17.x, leaving 1.16.x as it remains part of the GH Ubuntu 20.04 environment even though it is officially unsupported.  Go 1.15.x is dropped from the test matrix.

We also update the core libseccomp library to the latest v2.5.x release, which is v2.5.4 as of April 21, 2022.  The previous libseccomp versions are left untouched in the test matrix, even though they are no longer supported upstream.

[1] https://go.dev/doc/devel/release#policy

Signed-off-by: Paul Moore <paul@paul-moore.com>

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,8 +15,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        go-version: [1.15.x, 1.16.x, 1.17.x]
-        libseccomp: ["v2.3.3", "v2.4.3", "v2.5.2", "HEAD"]
+        go-version: [1.16.x, 1.17.x, 1.18.x]
+        libseccomp: ["v2.3.3", "v2.4.3", "v2.5.4", "HEAD"]
 
     steps:
 


pcmoore:
@kolyshkin it's not letting me add you as a reviewer yet, but your feedback would be appreciated.

drakenclimber:
> @kolyshkin it's not letting me add you as a reviewer yet, but your feedback would be appreciated.

I think this is because we haven't added him as a collaborator yet.  Not sure if this needs to be done on an org level (seccomp) or a repo level (seccomp/libseccomp-golang)

drakenclimber:
> I think this _can_ be done on the repo level.

I've added you as a collaborator, @kolyshkin.  Check your email from for the invite.

kolyshkin:
>  leaving 1.16.x as it remains part of the GH Ubuntu 20.04 environment even though
> it is officially unsupported. Go 1.15.x is dropped from the test matrix.

The fact that GH Ubuntu 20.04 is not really relevant, since in our CI jobs that require Go we never use the bundled Go version, always explicitly installing the one specified by the matrix.

In other words, the fact that Ubuntu comes with Go 1.16 should not stop us from dropping it from the CI matrix (or, in general, from the list of supported Go versions). OTOH we can be extra cautious and keep 1.16.x for the time being (say until Go 1.19 release or so).

In any case,

`Acked-by: Kir Kolyshkin <kolyshkin@gmail.com>`

kolyshkin (approved):

pcmoore:
> I think this is because we haven't added him as a collaborator yet. Not sure if this needs to be done on an org level (seccomp) or a repo level (seccomp/libseccomp-golang)

I should have been more clear, my comment was more along the lines of: "We haven't officially made Kir a maintainer yet, so I can't list him as reviewer in the GH UI, but I would still appreciate his feedback.".  See issue #87.

FWIW, I think it's okay to add @kolyshkin as a triage collaborator until we close out #87, and once we do we'll add him to the golang bindings admin group which should give him the ACLs he needs.

pcmoore:
Merged via f57e1d55ea18cbace22a9eefbd5e83417a2c1ace.

PR #93, opened 2022-05-06
RFE: add Kir to SECURITY.md

kolyshkin:
Related to https://github.com/seccomp/libseccomp-golang/issues/87


drakenclimber:
I'm totally good with this.  Thanks so much for all of the help, @kolyshkin and @pcmoore

PR #99, opened 2022-06-13
doc: init: add libseccomp version and TODO to comment

kolyshkin:
It was not clear from the commit 78c92cbc6966 what versions of
libseccomp require this kludge.

Looks like the issue in question was fixed by libseccomp commit
88afa50ff7c9787 ("api: force an API level update when necessary") which
made its way to libseccomp v2.5.0.

Amend the comment to tell the version, and add a TODO item to remove the
workaround.

### DIFF
--- a/seccomp.go
+++ b/seccomp.go
@@ -33,8 +33,9 @@ type VersionError struct {
 
 func init() {
 	// This forces the cgo libseccomp to initialize its internal API support state,
-	// which is necessary on older versions of libseccomp in order to work
+	// which is necessary on older versions of libseccomp (< 2.5.0) in order to work
 	// correctly.
+	// TODO: remove once libseccomp < v2.5.0 is not supported.
 	_, _ = getAPI()
 }
 


pcmoore (approved):
This looks fine to me, but FWIW, upstream no longer supports any libseccomp releases below v2.5.0 (technically we only support v2.5.4).

`Acked-by: Paul Moore <paul@paul-moore.com>`

seccomp.go
```diff
@@ -33,8 +33,9 @@ type VersionError struct {
 
 func init() {
 	// This forces the cgo libseccomp to initialize its internal API support state,
-	// which is necessary on older versions of libseccomp in order to work
+	// which is necessary on older versions of libseccomp (< 2.5.0) in order to work
 	// correctly.
+	// TODO: remove once libseccomp < v2.5.0 is not supported.
```
pcmoore, line 38: It's up to you @kolyshkin, but I would suggest creating a RFE issue in the issue tracker so this isn't forgotten :)

drakenclimber:
Pushed to main in commit 272261475b7985259982553c32266016bd5b8110.  Thanks all!

PR #103, opened 2022-09-02
RFE: ci: drop go 1.16, add go 1.19

kolyshkin:


pcmoore (approved):
The logic seems reasonable to me, thanks @kolyshkin.
```
```

seccomp.go
```diff
@@ -112,12 +110,15 @@ type ScmpNotifReq struct {
 //
 // ID:    notification ID (must match the corresponding ScmpNotifReq ID)
 // Error: must be 0 if no error occurred, or an error constant from package
-//        syscall (e.g., syscall.EPERM, etc). In the latter case, it's used
-//        as an error return from the syscall that created the notification.
+//
+//	syscall (e.g., syscall.EPERM, etc). In the latter case, it's used
+//	as an error return from the syscall that created the notification.
```
drakenclimber, line 115: This formatting feels weird to me, but I don't have a strong opinion.  Let's make the tool happy, I guess.

drakenclimber (approved):
I had one rather pointless comment, @kolyshkin; feel free to ignore it.  This looks good to me.  Thanks for the help.

seccomp.go
kolyshkin: It is indeed weird, and it is not rendered the way I thought it would. Let me take a closer look; for now I'm converting this PR to draft.

kolyshkin: Fixed, ready for review/merge.

The reason `gofumpt` did this is since a recent go release (1.18? I am not sure) indented lines are treated as code blocks, so gofumpt inserted an empty line and reformatted those "code blocks" using tab instead of spaces.

All this would make sense if those were really code blocks. They are not, and they are in fact misplaced, and thus are not rendered correctly (see e.g. https://pkg.go.dev/github.com/seccomp/libseccomp-golang#ScmpNotifData).

The fix is to move the docstrings to just before the IDs they describe, and remove the indentation. This is what the first commit now does.

drakenclimber (approved):
Looks good to me.  Thanks @kolyshkin 

pcmoore (approved):
Looks good to me, thanks.
`Acked-by: Paul Moore <paul@paul-moore.com>`

PR #105, opened 2023-02-08
Add Go 1.20 support

austinvazquez:
Adds Go 1.20 support

Signed-off-by: Austin Vazquez <macedonv@amazon.com>


austinvazquez:
Looks a little strange without https://github.com/seccomp/libseccomp-golang/pull/103, but figured I could rebase and resolve or that one can be updated to include this change. Whichever helps.

PR #110, opened 2024-08-14
CI bumps

kolyshkin:
Please see individual commits for details. This includes bump to Go 1.23.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -12,7 +12,7 @@ on:
 jobs:
 
   lint:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v4
       - name: install deps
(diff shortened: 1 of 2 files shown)


kolyshkin:
OK, I guess what happens here is actions/setup-go@v5 caches some Go files and then they are restored in an environment with different version of libseccomp installed, resulting in sporadic failures.

Let's try to add libseccomp version to cache key.

kolyshkin:
> Let's try to add libseccomp version to cache key.

It works, and I am seeing from the logs that cache is being reused.

That said, it will be an improvement to also cache compiled libseccomp, but that's a task for another PR.

.github/workflows/validate.yml
```diff
@@ -24,11 +24,11 @@ jobs:
           version: v1.60
 
   codespell:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
     steps:
     - uses: actions/checkout@v4
     - name: install deps
       # Version of codespell bundled with Ubuntu is way old, so use pip.
```
drakenclimber, line 31: Does v24.04 of Ubuntu still use a really old version of codespell?  If it has a current version, should we just switch to `apt install codespell`

kolyshkin: It doesn't now, but over time it will.

That reminds me of one thing I forgot here -- pin the codespell version (otherwise, when a new version is released, CI randomly breaks). Will do.

kolyshkin: added

drakenclimber:
Looks good to me.  Thanks, @kolyshkin 

timscharfenort8 (commented):
Good work und

PR #112, opened 2025-01-10
Added LoongArch architecture name

zhaixiaojuan:
When running the runc test, an error message is reported: "enosys_linux_test.go:179: unknown libseccomp architecture "loong64": could convert unrecognized string "loong64"". After checking, it was found that the error was caused by the missing "loong64" in the seccomp.go file.

### DIFF
--- a/seccomp.go
+++ b/seccomp.go
@@ -330,7 +330,7 @@ func GetArchFromString(arch string) (ScmpArch, error) {
 		return ArchPARISC64, nil
 	case "riscv64":
 		return ArchRISCV64, nil
-	case "loongarch64":
+	case "loong64", "loongarch64":
 		return ArchLOONGARCH64, nil
 	case "m68k":
 		return ArchM68K, nil


LocutusOfBorg:
this is already merged in main

seccomp.go
```diff
@@ -310,7 +310,7 @@ func GetArchFromString(arch string) (ScmpArch, error) {
 		return ArchPARISC64, nil
 	case "riscv64":
 		return ArchRISCV64, nil
-	case "loongarch64":
+	case "loong64", "loongarch64":
```
kolyshkin, line 313: From what I see, the official name is `loongarch64`. While I find the `arch` substring redundant, it is what it is. For example, see https://github.com/seccomp/libseccomp/blob/7db46d72f13c172b290818f624c2966bd0db5677/src/arch.c#L210 in which we only have `loongarch64` and not `loong64`.

kolyshkin (commented):
OTOH the official Go name is loong64. Hmm.

kolyshkin:
> OTOH the official Go name is loong64. Hmm.

It doesn't matter, really. Say GOARCH is `amd64` while seccomp's name is `x86_64`.

kolyshkin:
IOW, this library should support the same names as libseccomp.

zhaixiaojuan:
> > OTOH the official Go name is loong64. Hmm.
> 
> It doesn't matter, really. Say GOARCH is `amd64` while seccomp's name is `x86_64`.

As you said, GOARCH is `amd64`, and the name of seccomp is `x86_64`, so in libseccomp-golang/seccomp.go(https://github.com/seccomp/libseccomp-golang/blob/main/seccomp.go), the case branch adds x86_64 and amd64 to solve the name inconsistency problem:

`
switch strings.ToLower(arch) {
case "x86":
return ArchX86, nil
case "amd64", "x86-64", "x86_64", "x64":
return ArchAMD64, nil

`

For LoongArch, GOARCH is loong64 and seccomp is loongarch64. Similarly, we should include loong64 and loongarch64 in the case branch to solve the name inconsistency problem, runc is a good example, and this problem has blocked the runc test build upstream of debian.

zhaixiaojuan:
> this is already merged in main

Currently only loongarch64 is included, loong64 is missing:
`
	case "loongarch64":
		return ArchLOONGARCH64, nil
`

For this problem, the relevant patch has been merged into Debian in advance(https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1095452).

zhaixiaojuan:
@kolyshkin 
Hi, is there any latest progress on this issue?

zhaixiaojuan:
@kolyshkin 
Hi, I still hope that this problem can be fixed here, not just in the Debian upstream community. 
The code for fixing it in Debian upstream is as follows:
```
 const (
@@ -306,6 +308,8 @@ func GetArchFromString(arch string) (Scm
 		return ArchPARISC64, nil
 	case "riscv64":
 		return ArchRISCV64, nil
+	case "loongarch64", "loong64":
+		return ArchLOONGARCH64, nil
 	default:
 		return ArchInvalid, fmt.Errorf("cannot convert unrecognized string %q", arch)
 	}
```
Detailed patch link address:
https://bugs.debian.org/cgi-bin/bugreport.cgi?att=1;bug=1095322;filename=golang-github-seccomp-libseccomp-golang-add-loongarch64-support.patch;msg=5

kolyshkin:
@zhaixiaojuan please sign your commit.

Yes, I think `GetArchFromString` can and should recognize `GOARCH` values.

seccomp.go
kolyshkin: OTOH the official Go name is loong64 which I believe Go library should recognize.

kolyshkin (approved):
lgtm

PR #115, opened 2025-02-08
Add filter.Precompute

kolyshkin:
Adds support for seccomp_precompute, as well as a test case.


drakenclimber:
LGTM. Thanks, @kolyshkin

PR #121, opened 2025-03-14
ci: add all-done jobs

kolyshkin:
This is to simplify GH branch protection rules (to disable merging unless all the tests have succeeded).


kolyshkin:
drakenclimber @pcmoore PTAL

PR #123, opened 2025-04-08
ci: bump golangci-lint to v2

kolyshkin:
The configuration file format has changed.

Initially created via golangci-lint migrate, then edited to simplify.


drakenclimber (approved):
LGTM

PR #127, opened 2026-07-16
Fix FilterGetAttr returning EINVAL

kolyshkin:
This was initially reported to kubernetes[1] and fixed in runc[2] by ignoring EINVAL.

The original underlying issue though is when libseccomp-golang is compiled against an older version of libseccomp headers, the new flag attributes (such as WAITKILL) are set to 0 instead of a proper value, because for some reason _SCMP_FLTATR_MIN was used (starting from commit 4b175387).

As a result, seccomp_attr_set(3) returns EINVAL, as 0 is not a valid attribute value.

The fix is to provide the real values of the attributes so even when older headers are used we still have the correct values. NOTE we can't use #ifndef and instead rely on a headers version check because those values are part of an enum.

This bug eluded our tests because we always use the same libseccomp for compilation and testing.

[1]: https://github.com/kubernetes/kubernetes/issues/140039
[2]: https://github.com/opencontainers/runc/pull/5347

### DIFF
--- a/.github/actions/build-libseccomp/action.yml
+++ b/.github/actions/build-libseccomp/action.yml
@@ -1,0 +1,58 @@
+name: build libseccomp
+description: Build and install a specific version of libseccomp, and set up the environment to build and test against it.
+inputs:
+  version:
+    description: libseccomp version (git tag/branch) to build. Use vX.X.X to use a release tarball.
+    required: true
+
+runs:
+  using: composite
+  steps:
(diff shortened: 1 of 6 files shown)


kolyshkin:
> This bug eluded our tests because we always use the same libseccomp for compilation and testing.

Need to figure out what changes to do in CI to reveal this bug.

kolyshkin:
Closing in favor of #129.

PR #128, opened 2026-07-16
[DNM/test] ci: test against different libseccomp

kolyshkin:
This adds a CI step to run tests from the test binary which was built using older libseccomp version against the system bundled version as well as against latest libseccomp.

This reveals a bug fixed in #127, and two more:

1. When compiled against libseccomp 2.3.1 but used with [dynamically linked] libseccomp 2.4.0+, GetAPI always returns 0. This is being fixed by trying to dlsym the `seccomp_api_get` before returning EOPNOTSUPP.

2. When compiled against libseccomp < 2.5.0 but used with [dynamically linked] libseccomp 2.5.0+, `notifSupported` checks the _actual_ libseccomp version, not the one we compiled against, and thus mistakenly tells notification support is available, but the seccomp_notif_* stubs then return EOPNOTSUPP. This is being fixed by adding a compile time guard.

### DIFF
--- a/.github/actions/build-libseccomp/action.yml
+++ b/.github/actions/build-libseccomp/action.yml
@@ -1,0 +1,58 @@
+name: build libseccomp
+description: Build and install a specific version of libseccomp, and set up the environment to build and test against it.
+inputs:
+  version:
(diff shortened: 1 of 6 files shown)


kolyshkin:
This made its way into #129.

PR #129, opened 2026-07-17
Workarounds for different libseccomp version during runtime (vs compile time)

kolyshkin:
This started a week ago when I looked into https://github.com/kubernetes/kubernetes/issues/140039 and https://github.com/opencontainers/runc/pull/5347, and unfolded to be a much bigger problem.

When we use different libseccomp versions during compile and run time, all hell break loose due to some of the code assumptions being wrong. Example:

https://github.com/seccomp/libseccomp-golang/blob/3f07703264fb9b741cc707b57de80935e8994b4d/seccomp.go#L25-L29

In here, we use seccomp_precompute wrapper which returns an error for *compile-time* libseccomp < 2.6.0. Next, `checkAPI` and `checkVersion` checks are done against the *runtime* libseccomp, and thus if runtime libseccomp >= 2.6.0 we actually call this wrapper and get an error.

For this PR, I chose a conservative way to fix this: limit the version number (and the API level) to whatever was available during compile time. This way, the new functionality of newer libseccomp at runtime is not available (until you recompile against its headers), but there are no bad surprises either.

(A more radical approach would be to try to get most of the runtime libseccomp version, even if it's newer, but that needs `-ldl`, use of `dlsym` etc).

Initial CI improvements made while working on this are now moved to #130.

### DIFF
--- a/seccomp_compat.c
+++ b/seccomp_compat.c
@@ -1,72 +1,121 @@
-// Fallback implementations for libseccomp functionality that is missing in
-// older libseccomp headers/libraries. Declarations, and the version checks
-// that gate them, live in seccomp_compat.h.
+// Wrappers around libseccomp functionality that may be missing from the
+// run-time libseccomp library. The functions wrapped here are declared as weak
+// references (see seccomp_compat.h), meaning they resolve to NULL when the
+// run-time library is older than the version that added them, so they must not
+// be called directly.
 
 #include "seccomp_compat.h"
 
-#if SCMP_VER_MAJOR == 2 && SCMP_VER_MINOR < 4
+// The API level operations were added in libseccomp v2.4.0.
 
-// The libseccomp API level functions were added in v2.4.0.
-const unsigned int seccomp_api_get(void)
+unsigned int compat_api_get(void)
 {
-	// libseccomp-golang requires libseccomp v2.3.1, at a minimum, which
-	// supported API level 2. However, the kernel may not support API level
-	// 2 constructs which are the seccomp() system call and the TSYNC
-	// filter flag. Return the "reserved" value of 0 here to indicate that
-	// proper API level support is not available in libseccomp.
-	return 0;
+	// Return the "reserved" value of 0 to tell the caller that proper API
+	// level support is not available in libseccomp.
+	if (seccomp_api_get == NULL)
+		return 0;
+
+	return seccomp_api_get();
 }
 
-int seccomp_api_set(unsigned int level)
+int compat_api_set(unsigned int level)
 {
-	return -EOPNOTSUPP;
+	if (seccomp_api_set == NULL)
+		return -EOPNOTSUPP;
+
+	return seccomp_api_set(level);
 }
 
-#endif // < 2.4.0
 
+// The seccomp notify API was added in libseccomp v2.5.0.
 
-#if SCMP_VER_MAJOR == 2 && SCMP_VER_MINOR < 5
+int compat_notify_alloc(struct seccomp_notif **req, struct seccomp_notif_resp **resp)
+{
+	if (seccomp_notify_alloc == NULL)
+		return -EOPNOTSUPP;
 
-// The seccomp notify API functions were added in v2.5.0.
-
-int seccomp_notify_alloc(struct seccomp_notif **req, struct seccomp_notif_resp **resp) {
-	return -EOPNOTSUPP;
-}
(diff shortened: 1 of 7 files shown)


kolyshkin:
@drakenclimber @pcmoore can you please take a look? This fixes a real issue.

seccomp_compat.c
```diff
@@ -0,0 +1,72 @@
...
+// that gate them, live in seccomp_compat.h.
+
+#include "seccomp_compat.h"
+
+#if SCMP_VER_MAJOR == 2 && SCMP_VER_MINOR < 4
+
+// The libseccomp API level functions were added in v2.4.0.
+const unsigned int seccomp_api_get(void)
+{
+	// libseccomp-golang requires libseccomp v2.2.0, at a minimum, which
```
drakenclimber: nitpick - Did you mean v2.3.1 here?  In seccomp_compat.h, you explicitly state that v2.3.1 or higher is required

drakenclimber (approved):
Acked-by: Tom Hromatka (Oracle) <tom.hromatka@gmail.com>`

drakenclimber:
I also ran the patchset through Codex.  It came up with another error, but I'm not sure how I feel about it:

> This still breaks the compile-newer/run-older case for libseccomp 2.3.x. init() unconditionally invokes getAPI(), which directly references seccomp_api_get; that symbol was added in 2.4. A binary built with ≥2.4 headers and run with a supported 2.3.x library will abort on symbol resolution before checkVersion can return a compatibility error. Please add an indirection/fallback for the newer-headers + older-runtime case (and cover it in CI); seccomp_api_set needs the same treatment.

v2.4.0 of libseccomp came out in March 2019.  I admit that I would be tempted to only support v2.4.0 and newer, and avoid the problem of builds that don't have `seccomp_api_get()`.

Thoughts?

drakenclimber:
> I also ran the patchset through Codex. It came up with another error, but I'm not sure how I feel about it:
> 
> > This still breaks the compile-newer/run-older case for libseccomp 2.3.x. init() unconditionally invokes getAPI(), which directly references seccomp_api_get; that symbol was added in 2.4. A binary built with ≥2.4 headers and run with a supported 2.3.x library will abort on symbol resolution before checkVersion can return a compatibility error. Please add an indirection/fallback for the newer-headers + older-runtime case (and cover it in CI); seccomp_api_set needs the same treatment.
> 
> v2.4.0 of libseccomp came out in March 2019. I admit that I would be tempted to only support v2.4.0 and newer, and avoid the problem of builds that don't have `seccomp_api_get()`.
> 
> Thoughts?

With that said, one way to solve this (and still support v2.3.1) would be to use weak references:

```
/* Only when headers declare the real functions. */
#if SCMP_VER_MAJOR > 2 || (SCMP_VER_MAJOR == 2 && SCMP_VER_MINOR >= 4)
#pragma weak seccomp_api_get
#endif

unsigned int compat_seccomp_api_get(void)
{
#if SCMP_VER_MAJOR > 2 || (SCMP_VER_MAJOR == 2 && SCMP_VER_MINOR >= 4)
	if (seccomp_api_get != NULL)
		return seccomp_api_get();
#endif
	return 0; /* existing “API unsupported” convention */
}

...
```

seccomp_compat.c
kolyshkin: This is not a new comment, I just moved it around. At the time it was written, we required v2.2.0+.

I was hesitant to change the comment because if I do, the commit no longer moves the code around. Having said that, it's probably better to fix the issue in place as it raises questions.

Fixed now.

kolyshkin:
@drakenclimber thanks for suggestion (I tried weak symbols before but for some reasons they didn't work for me last time).

I've implemented weak symbols for newer functionality, and tests of various libseccomp build/runtime version combinations. The patchset looks a bit excessive (and somewhat limiting -- theoretically, we can use newer libseccomp functionality, but I am not aiming for this). Yet I believe the fixes cover all the corner cases now.

If you have time, please take another look.

kolyshkin:
@pcmoore @drakenclimber please take a look

drakenclimber:
Thanks for adding the weak symbol logic, @kolyshkin.  I realize it was a lot of work, but I think it is a good change.

The changes look good to me.

`Acked-by: Tom Hromatka (Oracle) <tom.hromatka@gmail.com>`

kolyshkin:
0.12.0 with all this is out the door now

PR #130, opened 2026-07-20
CI bumps, Makefile improvements

kolyshkin:
(This is the first half of #129 but can be reviewed/merged here for simplicity).

See individual commits for details.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -13,8 +13,8 @@ jobs:
   lint:
     runs-on: ubuntu-24.04
     steps:
-      - uses: actions/checkout@v6
-      - uses: actions/setup-go@v6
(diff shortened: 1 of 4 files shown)


.github/workflows/validate.yml
drakenclimber: nitpick - looks like we missed bumping this checkout to v7

drakenclimber (approved):
Looks good to me.  I found one small nitpick, but that's an easy fix.  Thanks, @kolyshkin.

`Acked-by: Tom Hromatka (Oracle) <tom.hromatka@gmail.com>`

.github/workflows/validate.yml
kolyshkin: Thanks! Fixed.

